PR #3, opened 2019-05-22
Implement minimal BOM generation from Gradle.

angrylogic:
Hi, I've taken a shot at completing out the implementation of BOM generation for Gradle.  I'm not terribly familiar with the Gradle APIs and there may be more effective ways of doing this work, but this PR appears to be functional.

There are a few changes compared to the original approach including:

**Generating a list of sub projects**
This code generates a list of sub projects which are built as dependencies and filters them out.  In my very limited Gradle experience these tend to be part of the same code base and thus not an external dependency that you would want to track.

**Resolve POM metadata**
In order to pull out the license and organizational details from the dependencies the plugin now asks Gradle to fetch the POM associated with each dependency and then parses it in order to extract these details.  If this operation fails the plugin continues without annotating the extra details.

This is using the Maven libraries to parse the POM files with code borrowed from the Maven plugin.  If this dependency is undesired we can probably drop it and parse the POM XML files directly.

The approach of fetching the POM metadata is inspired by one of the [gradle license plugins](https://github.com/hierynomus/license-gradle-plugin).

----

I would love to not have to maintain a vendored copy of these library so please let me know how I can help get this merged in and built.  

Thanks!

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -132,6 +132,11 @@
             <version>${gradle.version}</version>
             <scope>provided</scope>
         </dependency>
+        <dependency>
+            <groupId>org.apache.maven</groupId>
+            <artifactId>maven-core</artifactId>
+            <version>3.0.5</version>
+        </dependency>
     </dependencies>
 
     <build>
--- a/src/main/java/org/cyclonedx/gradle/CycloneDxTask.java
+++ b/src/main/java/org/cyclonedx/gradle/CycloneDxTask.java
@@ -18,23 +18,31 @@
 package org.cyclonedx.gradle;
 
 import org.apache.commons.io.FileUtils;
+import org.apache.maven.project.MavenProject;
 import org.cyclonedx.BomGenerator;
 import org.cyclonedx.BomParser;
 import org.cyclonedx.model.Component;
+import org.cyclonedx.model.License;
+import org.cyclonedx.util.BomUtils;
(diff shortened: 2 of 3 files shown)


angrylogic:
Hi @stevespringett, wanted to make sure this was on your radar.  The build failure here looks like a problem with the CI setup and I think is unrelated to this change.  Thanks!

stevespringett:
Thanks for pinging this ticket. I'll likely review this ticket tonight or tomorrow. Do you have any sample Gradle projects that you've tested with that are available on GitHub?

matteoturra:
Sono fuori dall'ufficio fino a 14/06/2019




Nota: Questa è una risposta automatizzata al messaggio  "Re:
[CycloneDX/cyclonedx-gradle-plugin] Implement minimal BOM generation from
Gradle. (#3)" inviata il 12/06/2019 05:02:34.

Questa è l'unica notifica che verrà ricevuta mentre la persona è assente.
Questa e-mail ed i suoi allegati sono ad uso esclusivo di coloro ai quali è indirizzata e potrebbe contenere informazioni riservate, la cui diffusione, copia e/o distribuzione è proibita. Qualora riceveste questo messaggio per errore, scusandoci per l'accaduto, Vi preghiamo di avvertirci immediatamente, distruggendo quanto ricevuto.

PR #8, opened 2019-07-26
Add control over skipped configurations

jskovjyskebankdk:
Allow clients to control which configurations to exclude when the BOM is generated.

At least with Gradle 5.4.1, we need to exclude "compileClasspath" and "testCompileClasspath" in addition to those in the task's list.
This also allows us to exclude custom compile-time configurations which should not affect the BOM.


stevespringett:
Great. Thanks for the PR.

stevespringett:
v1.1.0 was released to central and will be available shortly.

PR #12, opened 2019-08-05
Issue6

stefanneuhaus:
- Applied the plugin using the normal way as proposed by the Gradle Team; see [Using your plugin in another project](https://docs.gradle.org/current/userguide/custom_plugins.html#sec:using_your_plugin_in_another_project).
- Updated documentation


stevespringett:
Thanks for the PR. Does this PR work with older versions of Gradle or just the latest?

I don't use gradle and find it really difficult to use/test.

stefanneuhaus:
It also works with reasonably old versions of Gradle. Tested with:
- ✅ 5.5.1 (2019-07-10)
- ✅ 5.0 (2018-11-26)
- ✅ 4.0 (2017-06-14)
- ✅ 3.0 (2016-08-15)
- ✅ 2.0 (2014-07-01)
- ❌ 1.0 (2012-06-12)

[SDKMAN!](https://sdkman.io/) can be used to easily switch Gradle versions like:
```sdk use gradle 5.5.1```

**Comment by sschuberth** [comment]
Please note that using `buildscript {}` instead of `plugins {}` actually is *not* "the normal way as proposed by the Gradle Team", but the [legacy](https://docs.gradle.org/5.6.1/userguide/plugins.html#sec:old_plugin_application) way of applying plugins.

The current way of applying plugins is by using the plugins DSL, and using [plugin resolution rules](https://docs.gradle.org/5.6.1/userguide/plugins.html#sec:plugin_resolution_rules) to map plugin ids to arbitrary Maven coordinates, just like the [current README](https://github.com/CycloneDX/cyclonedx-gradle-plugin/blob/ef9f797883c7cc171a37ca5a4bc522db878d1386/README.md) already does.

**Comment by matteoturra** [comment]
Sono fuori dall'ufficio fino a 09/09/2019




Nota: Questa è una risposta automatizzata al messaggio  "Re:
[CycloneDX/cyclonedx-gradle-plugin] Issue6 (#12)" inviata il 02/09/2019
18:06:12.

Questa è l'unica notifica che verrà ricevuta mentre la persona è assente.
Questa e-mail ed i suoi allegati sono ad uso esclusivo di coloro ai quali è indirizzata e potrebbe contenere informazioni riservate, la cui diffusione, copia e/o distribuzione è proibita. Qualora riceveste questo messaggio per errore, scusandoci per l'accaduto, Vi preghiamo di avvertirci immediatamente, distruggendo quanto ricevuto.

**Comment by lislei** [comment]
Using the current recommended way of declaring plugin usage in Gradle then I have the following in `settings.gradle`
```
pluginManagement {
    plugins {
        id 'org.cyclonedx.bom' version '1.4.1'
    }
    repositories {
        mavenCentral()
    }
}
```
and this in `build.gradle`
```
plugins {
    id 'org.cyclonedx.bom'
}
```

Perhaps a link on how to declare repository for the custom plugin dependencies would be useful. 
- https://docs.gradle.org/current/userguide/plugins.html#sec:custom_plugin_repositories

For the legacy `buildscript {}` example then maybe a link to the Gradle Plugin Portal page would suffice?
- https://plugins.gradle.org/plugin/org.cyclonedx.bom

glefloch:
Thanks for this PR @stefanneuhaus, lately we updated the documentation. I'm closing this as we pushed similar updates in different pr.

PR #15, opened 2019-09-03
README: Simplify instructions to apply the plugin

sschuberth:
"apply true" is the default anyway, and it is enough to mention the
version only once as part of the pluginManagement block.


stevespringett:
Thanks for the PR

PR #17, opened 2019-09-03
CycloneDxPlugin: Set a group for the task to make it visible

### LINKED ISSUE #14: cyclonedxBom task should have a task group assigned
Otherwise `./gradlew tasks` [does not list the task](https://stackoverflow.com/a/23598928/1127485), only `./gradlew tasks --all` does (and then currently in the "Other tasks" group).

sschuberth:
This makes the task show up in "./gradlew tasks". Previously "./gradlew
tasks --all" was needed.

Fixes #14.


stevespringett:
Thanks for the PR

PR #31, opened 2019-11-03
Make the POM error more useful by emitting the purl.

llamahunter:
Currently it just emits the class object reference, which is useless
when trying to track down which POM is missing.

### DIFF
--- a/src/main/java/org/cyclonedx/gradle/CycloneDxTask.java
+++ b/src/main/java/org/cyclonedx/gradle/CycloneDxTask.java
@@ -183,9 +183,9 @@ public class CycloneDxTask extends DefaultTask {
             final File pomFile = pomCfg.resolve().stream().findFirst().orElse(null);
             project = mavenHelper.readPom(pomFile);
         } catch(IOException err) {
-            getLogger().error("Unable to resolve POM for " + component + ": " + err);
+            getLogger().error("Unable to resolve POM for " + component.getPurl() + ": " + err);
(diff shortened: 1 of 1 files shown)


src/main/java/org/cyclonedx/gradle/CycloneDxTask.java
```diff
@@ -264,6 +264,7 @@ private boolean shouldSkipConfiguration(Configuration configuration) {
 
     /**
      * Ported from Maven plugin.
+     * @param components The CycloneDX components extracted from gradle dependencies
```
llamahunter: The build was complaining about this missing doc comment, so added one.

src/main/java/org/cyclonedx/gradle/CycloneDxTask.java
```diff
@@ -183,9 +183,9 @@ private void augmentComponentMetadata(Component component, String dependencyName
             final File pomFile = pomCfg.resolve().stream().findFirst().orElse(null);
             project = mavenHelper.readPom(pomFile);
         } catch(IOException err) {
-            getLogger().error("Unable to resolve POM for " + component + ": " + err);
+            getLogger().error("Unable to resolve POM for " + component.getPurl() + ": " + err);
```
llamahunter: at least with this there will be a hope of tracking down the offender

stevespringett:
Thanks for the PR Richard.

PR #36, opened 2019-11-18
Add build.gradle to support Gradle Plugin Portal publishing.

llamahunter:
A simple build.gradle that supports building the plugin with the gradle
plugin helper (performing extra validation checks), and publishing to
the Gradle Plugin Portal (GPP), as well as local maven for testing.
Needs manual maintenance of the 'version' field to keep it in sync with
the master pom.xml file.  GPP publishing requires an account on the
plugins.gradle.org website.


llamahunter:
See the README-GradlePluginPortal.md for use of the build.gradle for doing publishing to the GPP.  Note that it requires setting up an account on plugins.gradle.com to get API keys.  I did a test publish using the 'name.richardlee.cyclonedx.bom' id, so hopefully you won't have any ownership collision problems when eventually publishing it to the correct 'org.cyclonedx.bom' id.  If there are ownership issues of some sort, there's a process to sort that out at https://github.com/gradle/plugin-portal-requests

build.gradle
```diff
@@ -0,0 +1,47 @@
+
+plugins {
+    id 'java-gradle-plugin'
+    id 'com.gradle.plugin-publish' version '0.10.1'
```
llamahunter: the gradle plugin portal publisher

build.gradle
```diff
@@ -0,0 +1,47 @@
+
+plugins {
+    id 'java-gradle-plugin'
+    id 'com.gradle.plugin-publish' version '0.10.1' 
+    id 'maven-publish'
+}
+
+repositories {
+    jcenter()
```
llamahunter: the gradle plugin docs recommend using jcenter() as the repository, as that's what gradle will use by default at runtime to resolve plugins.  It's better for downstream users if jcenter() 'just works' without having to add a bunch of buildscript repositories, so better to build your plugin a jcenter() to make sure there aren't unexpected issues that can't be resolved during development.

build.gradle
```diff
@@ -0,0 +1,47 @@
...
+}
+
+dependencies {
+    compile(group: 'org.cyclonedx', name: 'cyclonedx-core-java', version: '2.1.1') {
+        // gradle-api already includes an slf4j binding
+        exclude group: 'org.apache.logging.log4j', module: 'log4j-slf4j-impl'
```
llamahunter: Not sure why cyclonedx-core-java has an _implementation_ for slf4j in it.  Generally bad practice for libraries to include implementations of the logger.  Those should be bound by the runtime environment, i.e. gradle, in this case.

build.gradle
```diff
@@ -0,0 +1,47 @@
...
+    }
+    compile 'commons-codec:commons-codec:1.13'
+    compile 'commons-io:commons-io:2.6'
+    compile 'org.apache.maven:maven-core:3.5.0'
+}
+
+group = 'com.cyclonedx'
+version = '1.1.2-SNAPSHOT'
```
llamahunter: VERY IMPORTANT  This 'version' field needs to be manually maintained in sync with pom.xml

build.gradle
```diff
@@ -0,0 +1,47 @@
...
+targetCompatibility = JavaVersion.VERSION_1_8
+
+tasks.withType(JavaCompile) {
+    options.encoding = 'UTF-8'
+}
+
+pluginBundle {
+    website = 'http://cyclonedx.org'
+    vcsUrl = 'https://github.com/CycloneDX/cyclonedx-gradle-plugin.git'
+    tags = [ 'cyclonedx', 'dependency', 'dependencies', 'owasp', 'bom' ]
```
llamahunter: I made up some tags.  Feel free to edit as you see fit

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -0,0 +1,5 @@
+distributionBase=GRADLE_USER_HOME
+distributionPath=wrapper/dists
+distributionUrl=https\://services.gradle.org/distributions/gradle-5.6.4-bin.zip
+zipStoreBase=GRADLE_USER_HOME
+zipStorePath=wrapper/dists
```
llamahunter: gradle wrapper configuration

gradlew
```diff
@@ -0,0 +1,188 @@
+#!/usr/bin/env sh
```
llamahunter: this whole thing is gradle wrapper boilerplate

gradlew.bat
```diff
@@ -0,0 +1,100 @@
+@rem
```
llamahunter: also all gradle wrapper boilerplate (for windoze)

settings.gradle
```diff
@@ -0,0 +1,5 @@
+/*
+ * This file was generated by the Gradle 'init' task.
+ */
+
+rootProject.name = 'cyclonedx-gradle-plugin'
```
llamahunter: By default, gradle will use the directory name as the project name.  Sometimes this causes weird things to happen if people map the directory name to something custom.  This sets the project name explicitly to avoid such problems.

src/main/java/org/cyclonedx/gradle/CycloneDxTask.java
```diff
@@ -76,6 +77,7 @@
     private boolean skip;
     private final List<String> skipConfigs = new ArrayList<>();
 
+    @Input
```
llamahunter: The gradle plugin helper did not like that this field did not declare it was an input for the plugin.  There are other fields that could be also specified as inputs (i.e. buildDir), but they need getters to be treated as such.

llamahunter:
@stevespringett unsure how to assign you as a reviewer, but wanted to make sure you saw this.

build.gradle
stevespringett: Agreed. I think its inherited due to the spdx-tools dependency, which is a bit of a mess.

build.gradle
stevespringett: I'll likely add this to the pom so that its automatic

PR #43, opened 2020-01-06
Resolution performance improvements

mwfriedm:
Cache the maven project models and the artifact file hashes, so they don't get reevaluated for the same artifacts.  A given artifact may be a dependency of multiple configurations in multiple subprojects.  For the same artifact, the maven model and the files hashes aren't going to change though!

Parallelize the resolution of POMs for additional metadata.  The detached configurations for these are generated on the fly, and don't have any dependencies between each other.  Gradle will happily resolve them in parallel.

These changes yield an order of magnitude improvement in runtime on my large, multiproject build.

PR #63, opened 2020-04-07
Avoid fetching the same dependency twice

dwursteisen:
Fetching the information from the pom.xml can took some time. 
Already analysed dependencies are saved. If the same dependency needs
to be analysed again, it's skipped.

In our project, `./gradlew cyclonedxBom` goes from 12 minutes to 2 minutes with this fix.

### DIFF
--- a/src/main/java/org/cyclonedx/gradle/CycloneDxTask.java
+++ b/src/main/java/org/cyclonedx/gradle/CycloneDxTask.java
@@ -111,7 +111,7 @@ public class CycloneDxTask extends DefaultTask {
         }
         logParameters();
         getLogger().info(MESSAGE_RESOLVING_DEPS);
-        final Set<String> builtDependencies = getProject()
+        final Set<String> analysedDependencies = getProject()
                 .getRootProject()
                 .getSubprojects()
                 .stream()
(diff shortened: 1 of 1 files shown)


stevespringett:
Thanks for the PR. I think this is a subset of the functionality available in #43

dwursteisen:
Yes.

It's the same idea. I just try to change the fewest lines possible.

mamuf:
One issue with this PR is that all the analysed dependencies of given module+config combo are then logged, but this fix "breaks" the logged information by only including the dependencies that were actually processed in any given step. And so if by any chance the order of submodules changes, the logged output for each module+config combo will not be the same.

By all means, I want this performance fix, but it needs a bit more work so that the logged information stays the same (i.e. dependencies effectively processed, even when they were skipped for the particular module+config). But perhaps the mentioned PR solves this (didn't study it that well).

dwursteisen:
Hello

Thanks for your reply. 

Can you be more explicit for this part, I'm not sure to get your point:

> but it needs a bit more work so that the logged information stays the same (i.e. dependencies effectively processed, even when they were skipped for the particular module+config). But perhaps the mentioned PR solves this (didn't study it that well).

- By skipped, you mean spiked by my fix or skipped by gradle because of a configuration parameter or what so ever ? (like a custom configuration?)
- By logged information, you mean the gathering of all information that will be produced by a build? (and not an text output of the plugin that I might have miss)

If I'm understanding your thought, it would be better, instead of "skipping" an already analysis dependencies, to put it in a cache so the output will always the same no matter the order of the analysis?

Regards

mamuf:
@dwursteisen Hi.

By "skipped" I mean by your fix, because given dependency was already processed as part of perhaps another module or another configuration in the same module. So yes, basically what you say at the end.

If you look at the line 144 in your patch, there's the logging I am talking about. With the original "unoptimized" code, every dependency for any given module+config combo is logged there as part of the process. Your fix "skips" the processing of already processed dependencies and in the end they are not logged as part of that module+config. So yes, putting it in a "cache" or something like that would be the proper solution in my eyes: The performance would improve massively and both the produced BOM file and logged output would still be the same.

dwursteisen:
👍  Thanks for your explanation. 

Regarding those elements, I'll close this Pull request. 
I'll look again at the plugin code if I can implement a solution that would more use a cache instead (if the other MR doesn't implement it)

Regards.

PR #88, opened 2020-12-28
Implement Caching and use streaming interfaces

bdellegrazie:
This is a rebased and squashed version of @mwfriedm earlier work in #43. The work is @mwfriedm's, not mine.
There's always the chance I made a mistake while doing this so any comments / reviews appreciated.

In order to avoid Illegal state exceptions from Gradle I have removed the parallelisation (`parallelStream` -> `stream`) but this is trivially easy to reverse once those issues are resolved. It appears to be a threading model issue when integrating with Gradle
(gradle plugins need to tell Gradle about worker threads).

This provides a reduction from 7 minutes down to just under 1 on our project.


mwfriedm:
This looks good to me.  Thank you for picking this up!

stevespringett:
Huge thanks to you both.

PR #110, opened 2022-02-07
Leverage lazy task configuration

glefloch:
this leverage gradle lazy api to avoid creating the task if the task is not executed.


glefloch:
@stevespringett it looks like the maven `pom.xml` does reference an old version of gradle should we update it ?

stevespringett:
What is the minimal viable version of Gradle that needs to be supported?

How will that affect adopters in terms of marketshare in use?

PR #113, opened 2022-02-10
Metadata filtering for tool in sbom

lislei:
Hello Steve Springett.

I noticed the gradle plugin didn't fill inn the metadata of the tool producing the SBOM.

This pull request adds filtering capability to the gradle build. 
Also, the commons-io dependency is replaced with use of the standard JDK-api. 

I've successfully verified the contents of the property file as well as outputted XML and JSON files. 
This produces similar data as what the maven plugin does. 

What I'm uncertain of is the format of the value for "timestamp" as it is left unexpanded by the maven plugin. 
The [SBOM 1.4 schema ](https://github.com/CycloneDX/specification/blob/master/schema/bom-1.4.xsd#L149) doesn't specify this explicitly.
Is this field unused and to be removed or should it be formatted in a certain way?

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,4 +1,3 @@
-
 plugins {
     id 'java-gradle-plugin'
     id 'com.gradle.plugin-publish' version '0.12.0'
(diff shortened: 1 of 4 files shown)


lislei:
My test produced this XML
```
  <metadata>
    <timestamp>2022-02-10T16:25:56Z</timestamp>
    <tools>
      <tool>
        <vendor>CycloneDX</vendor>
        <name>CycloneDX Gradle plugin</name>
        <version>1.5.0-SNAPSHOT</version>
      </tool>
    </tools>
```
.. and the content of the plugin property file is:
```
# Automatically populated by Gradle build - do not modify
vendor=CycloneDX
name=CycloneDX Gradle plugin
version=1.5.0-SNAPSHOT
timestamp=1644510352444
```

jeremylong:
@skhokhlov this PR can be closed as the change https://github.com/CycloneDX/cyclonedx-gradle-plugin/pull/532/files#diff-68e7e71a19aba289bf5bddffd63f21079a214e0efff535a3caca47cefe33c22dR127 populates the tools section.

Issue #95, opened 2021-06-30
Is it possible to define inputs and outputs for the Gradle Task so that the BOM is only newly created when the dependencies have changed?


gebhardt:
Generating the BOM takes a while, and it would be really nice, if it is only generated, if the dependencies have changed.
If they are unchanged, the result should be taken from the BOM.

stevespringett:
I'm not sure it would help performance that much. The hashes of all components would still need to be generated as the jars residing in .m2 and .gradle could change (maliciously, corruption, etc) without the stated dependencies in gradle changing. There's also network considerations. If resolving from a repo, there's usually no guarantee that a build system will retrieve the same dependency from the same mirror. In the event one mirror gets compromised and others do not, relying solely on previous hash values would lead to a false sense of security.

I'm open to proposals that would not have an adverse impact on security.

PR #117, opened 2022-03-09
Declare output directory

sal0max:
This pull request adds a method `getOutputDir`, with an `@OutputDirectory` annotation.

This is considered good practice. See
- https://docs.gradle.org/current/userguide/more_about_tasks.html#sec:up_to_date_checks
- https://docs.gradle.org/current/javadoc/org/gradle/api/tasks/OutputDirectory.html

> As part of incremental build, Gradle tests whether any of the task inputs or outputs has changed since the last build. If they haven’t, Gradle can consider the task up to date and therefore skip executing its actions. Also note that incremental build won’t work unless a task has at least one task output, although tasks usually have at least one input as well.

This is important for two reasons:
1. As mentioned above, `OutputDirectory` is needed for incremental builds, to detect when a task is still up-to-date.
2. This also allows the following:

---

Currently, to work with the generated `bom.xml` and `bom.json` files, one has to address them by directory, like this
```groovy
task myCopy(type: Copy) {
    dependsOn(cyclonedxBom)

    from("$buildDir/reports")
    into("$projectDir/copy")
}
```

When an output directory is declared one can streamline the function like this:
```groovy
task myCopy2(type: Copy) {
    from(cyclonedxBom)
    into("$projectDir/copy")
}
```

---

Maybe also solves https://github.com/CycloneDX/cyclonedx-gradle-plugin/issues/95

### DIFF
--- a/src/main/java/org/cyclonedx/gradle/CycloneDxTask.java
+++ b/src/main/java/org/cyclonedx/gradle/CycloneDxTask.java
@@ -47,6 +47,7 @@ import org.gradle.api.artifacts.ModuleVersionIdentifier;
 import org.gradle.api.artifacts.ResolvedArtifact;
 import org.gradle.api.artifacts.ResolvedConfiguration;
 import org.gradle.api.tasks.Input;
+import org.gradle.api.tasks.OutputDirectory;
 import org.gradle.api.tasks.TaskAction;
 
 import javax.xml.parsers.ParserConfigurationException;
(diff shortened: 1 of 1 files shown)


glefloch:
Thanks for this @sal0max I just push #136 which introduce a new destination output directory and I added the `@OutputDirectory` annotation on top of it.

sal0max:
Alright, thanks a lot!

PR #128, opened 2022-05-04
Fix plugin import in docs

glefloch:
This branch fix the local / remote plugin publication by removing a duplicate file. 
I also simplified the documentation on how to import the plugin and tested it with the project and in the test project. 

@stevespringett once, merge, you should be able to publish the version `1.5.1` of the plugin using the `cdx-automation` account as mentionned here https://github.com/gradle/plugin-portal-requests/issues/135


stevespringett:
Huge thanks Guillaume. I think the only thing missing at this point, is a GH action that performs the publishing.

Just added https://github.com/CycloneDX/cyclonedx-gradle-plugin/blob/master/.github/workflows/gradle-publish.yml. Need to test.

PR #129, opened 2022-05-04
Initialize dependabot configuration for gradle dependencies

glefloch:
This initialize `dependabot` configuration, it could be useful to get dependency update.

@stevespringett could we enable it on the repository ?


stevespringett:
Absolutely. Thank you.

PR #136, opened 2022-05-06
Introduce destination property allowing to configure output

### LINKED ISSUE #62: How to change the generated BOM location for gradle plugin
I want to generate into ./build/reports/cyclondx/bom.xml. Is it possible in gradle plugin? Thanks

### LINKED ISSUE #95: Is it possible to define inputs and outputs for the Gradle Task so that the BOM is only newly created when the dependencies have changed?

glefloch:
This introduce a new property allowing the user to set a destination. 

This also sets the destination as `@OutputDirectory` making the task `UP-TO-DATE` if nothing changed.

This also introduce some integration to validate outputs directories.

close #62 
close #95

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,6 +50,8 @@ cyclonedxBom {
     projectType = "application"
     // Specified the version of the CycloneDX specification to use. Defaults to 1.2.
     schemaVersion = "1.2"
+    // Boms destination directory (defaults to build/reports)
+    destination = file("build/reports")
 }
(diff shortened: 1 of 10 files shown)


stevespringett (approved):
LGTM

PR #142, opened 2022-05-12
Declare all task input as settable property

### LINKED ISSUE #138: [Feature Request] Full Gradle Kotlin Script support
### Description

While this plugin already partially works with a Project with `build.gradle.kts`, its options are limited in comparison to the Groovy based version.

Specifically, you can set only the `includeConfigs` and `skipConfigs`, but not the `projectType` nor the `schemaVersion`.

While the most relevant options can be set it would be great if Kotlin Script based projects would also get full config support.

### What is currently possible

`build.gradle.kts`
```kotlin
//...
task.withType<CycloneDxTask> {
  setIncludeConfigs(listOf("runtimeClasspath"))
  setSkipConfigs(listOf("compileClasspath", "testCompileClasspath"))
}
//...
```

### What would be ideally possible

Fully supported configuration of all options of the plugin within `build.gradle.kts` (+ added examples in the docs)

glefloch:
This declare all input as task property. This can then be set by both groovy and kotlin DSL. 

I removed the `skip` property as gradle provide the `-x` flag to skip a task and the `skip` property was not documented.

close #138

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,11 +23,6 @@ __Output CycloneDX Generation Info:__
 gradle cyclonedxBom -info
 ```
 
-__Exclude BOM Serial Number:__
-```bash
-gradle cyclonedxBom -Pcyclonedx.includeBomSerialNumber=false
-```
-
 __build.gradle__ (excerpt)
 ```groovy
 plugins {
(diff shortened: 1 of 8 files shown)


stevespringett (approved):
LGTM

PR #144, opened 2022-05-18
Populating plugin.properties on build to address #119

### LINKED ISSUE #119: metadata does not get templated
Using version 1.5.0 of this plugin, the `metadata` element of the generated sbom file always looks like this:

```
  "metadata" : {
    "timestamp" : "2022-03-28T18:16:01Z",
    "tools" : [
      {
        "vendor" : "${project.organization.name}",
        "name" : "${project.name}",
        "version" : "${project.version}"
      }
    ],

```

According to the source code, these attributes should be read from https://github.com/CycloneDX/cyclonedx-gradle-plugin/blob/master/src/main/resources/plugin.properties, but this template file doesn't seem to be rendered correctly.

ThomGeG:
While integrating the Gradle plugin into some of my company's projects, I noticed the `plugin.properties` was no longer being updated since the move away from using Maven. This PR should close #119, which already reports the problem this causes with the BOM meta-data.

Let me know if there's a better way of doing this, as my Gradle isn't strong.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@
 *~
 /build/
 /.gradle/
+/src/main/resources/plugin.properties
--- a/build.gradle
+++ b/build.gradle
@@ -30,8 +30,16 @@ test {
     useJUnit()
 }
 
+project.ext.organization = 'CycloneDX'
 group = 'org.cyclonedx'
 version = '1.5.1'
+
+// populate properties into plugin.properties at build time
+ant.propertyfile(file: "src/main/resources/plugin.properties") {
+  entry(key: "vendor", value: project.ext.organization)
+  entry(key: "name", value: project.name)
+  entry(key: "version", value: project.version)
+}
 
 sourceCompatibility = JavaVersion.VERSION_1_8
 targetCompatibility = JavaVersion.VERSION_1_8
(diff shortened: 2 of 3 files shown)


glefloch (commented):
Thanks for this @ThomGeG. I think we can remove the property file and extract property as constant directly in the task

glefloch:
The `group` is not use so we can remove it from the file, the `vendor` is already a constant in `build.gradle` I think we can move it to the task. The `version` could I think be find in the `META_INF/manifest.mf` file generated in the jar. That would avoid having to do custom things in the build.

ThomGeG:
I'm not sure it's possible to fetch these properties as a part of the `CycloneDxTask.java` from the `build.gradle` at run time is it? That's the kind of thing I was looking for online when making the original changes, but I couldn't find a way of actually doing it.

Decompiling the latest release, it looks like it'd have something to do with how these properties don't actually get compiled anywhere. The manifest doesn't actually contain anything meaningful and the only other properties file I can see is something internal to gradle and how it [finds the implementing class](https://docs.gradle.org/current/userguide/java_gradle_plugin.html#java_gradle_plugin):
image
image

The manifest probably should contain the version (and a timestamp of when it was built), but from what I've heard it's rather complicated to read values off it at run time. It would seem the simplest solution is to build our own properties at compile time like we currently are and then read from it.

glefloch:
Right let's keep it this way.

glefloch (approved):

PR #145, opened 2022-05-18
Add configuration for outputName

### LINKED ISSUE #143: Allow configuration of file names for BOMs
One of the things the [cyclonedx-maven-plugin](https://github.com/CycloneDX/cyclonedx-maven-plugin) lets you do is specify the filename for the BOMs being produced via the `outputName` configuration. The gradle plugin meanwhile just uses the same static name of `bom` for both the XML and JSON files.

https://github.com/CycloneDX/cyclonedx-gradle-plugin/blob/cb1fc941db17bc07cbf05ed46a1b4fa6b1dc61e0/src/main/java/org/cyclonedx/gradle/CycloneDxTask.java#L457

https://github.com/CycloneDX/cyclonedx-gradle-plugin/blob/cb1fc941db17bc07cbf05ed46a1b4fa6b1dc61e0/src/main/java/org/cyclonedx/gradle/CycloneDxTask.java#L439

It would be convenient for those that need it to add another input to allow configuration of these file names. If I have the time, I'll try throwing together a PR in the coming days to add it if nobody jumps in sooner

ThomGeG:
Resolves #143 by adding the same `outputName` configuration from the Maven plugin to the Gradle plugin.

I've also added some additional parameters to the logging that were missing around skip/include configs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,7 +30,7 @@ plugins {
 }
 ```
 
-Once a BOM is generated, it will reside at `./build/reports/bom.xml` and `./build/reports/bom.json`
+Once a BOM is generated, by default it will reside at `./build/reports/bom.xml` and `./build/reports/bom.json`
 
 
 __Configuration:__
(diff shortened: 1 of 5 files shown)


glefloch:
Thanks @ThomGeG, could you squash your commit into one?

glefloch (approved):

glefloch (approved):
Thanks!

PR #146, opened 2022-05-20
Fixing copy-paste error in README.md

ThomGeG:
Looks like when the README was made it was copy-pasted from the NPM module's README and not updated.


glefloch (approved):
Nice catch !

Issue #107, opened 2022-01-26
dependencies with ISO-8859-1 encoding of their POM are skipped


hedtke:
When I use the plugin on the following gradle project

```
plugins {
    id 'org.cyclonedx.bom' version '1.4.1'
    id 'com.github.johnrengelman.shadow' version '7.1.2'
    id 'java'
}

repositories {
    mavenCentral()
}

dependencies {
    testImplementation "org.junit.platform:junit-platform-runner:1.8.2"
}
```

I get the message

```
An error occurred attempting to read POM
org.codehaus.plexus.util.xml.pull.XmlPullParserException: UTF-8 BOM plus xml decl of ISO-8859-1 is incompatible (position: START_DOCUMENT seen <?xml version="1.0" encoding="ISO-8859-1"... @1:41) 
	at org.codehaus.plexus.util.xml.pull.MXParser.parseXmlDeclWithVersion(MXParser.java:3366)
	at org.codehaus.plexus.util.xml.pull.MXParser.parseXmlDecl(MXParser.java:3288)
	at org.codehaus.plexus.util.xml.pull.MXParser.parsePI(MXParser.java:3140)
	at org.codehaus.plexus.util.xml.pull.MXParser.parseProlog(MXParser.java:1828)
	at org.codehaus.plexus.util.xml.pull.MXParser.nextImpl(MXParser.java:1757)
	at org.codehaus.plexus.util.xml.pull.MXParser.next(MXParser.java:1375)
	at org.apache.maven.model.io.xpp3.MavenXpp3Reader.read(MavenXpp3Reader.java:3856)
	at org.apache.maven.model.io.xpp3.MavenXpp3Reader.read(MavenXpp3Reader.java:595)
	at org.apache.maven.model.io.xpp3.MavenXpp3Reader.read(MavenXpp3Reader.java:609)
	at org.cyclonedx.gradle.MavenHelper.readPom(MavenHelper.java:298)
	at org.cyclonedx.gradle.MavenHelper.readPom(MavenHelper.java:285)
	at org.cyclonedx.gradle.CycloneDxTask.getResolvedMavenProject(CycloneDxTask.java:222)
	at org.cyclonedx.gradle.CycloneDxTask.augmentComponentMetadata(CycloneDxTask.java:234)
	at org.cyclonedx.gradle.CycloneDxTask.lambda$null$3(CycloneDxTask.java:167)
[...]

PR #152, opened 2022-05-25
Use ReaderFactory.newXmlReader() instead of ISR in MavenHelper

fgunbin:
plexus-utils's MXParser obtains encoding from XmlReader (see https://github.com/codehaus-plexus/plexus-utils/issues/163) or from ISR.
In our case MavenHelper creates ISR, and it then reports "default" encoding, which most probably will be UTF-8.

When the pom file contains encoding declaration with something else, like ISO-8859-1 in https://repo1.maven.org/maven2/org/apache/commons/commons-parent/48/commons-parent-48.pom, this results in the error:

```
An error occurred attempting to read POM
org.codehaus.plexus.util.xml.pull.XmlPullParserException: UTF-8 BOM plus xml decl of iso-8859-1 is incompatible (position: START_DOCUMENT seen <?xml version="1.0" encoding="iso-8859-1"... @1:41)
	at org.codehaus.plexus.util.xml.pull.MXParser.parseXmlDeclWithVersion(MXParser.java:3366)
	at org.codehaus.plexus.util.xml.pull.MXParser.parseXmlDecl(MXParser.java:3288)
```

This PR switches to using plexus-utils's ReaderFactory, which "knows better".  This eliminates the above error.

### DIFF
--- a/src/main/java/org/cyclonedx/gradle/MavenHelper.java
+++ b/src/main/java/org/cyclonedx/gradle/MavenHelper.java
@@ -23,6 +23,7 @@ import java.io.FileInputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.InputStreamReader;
+import java.io.Reader;
 import java.net.URI;
 import java.net.URISyntaxException;
 import java.util.List;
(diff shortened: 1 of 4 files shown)


fgunbin:
Should fix #107

glefloch:
Thanks for this @fgunbin. Could you add a test that assert no more exceptions are present in logs?

fgunbin:
@glefloch Added, please check.

glefloch (approved):
Great thanks!

fgunbin:
@glefloch Thank you!

PR #153, opened 2022-05-28
fix: CycloneDx schema version to support 1.4

### LINKED ISSUE #154: Plugin 1.6.0 does not support CycloneDX Schema 1.4
### Description

Tried out the new Plugin version the other day and noticed that it does not support CycloneDX schema version 1.4, despite listing it as the default for the new version.
After attempting to manually set it to 1.4 and still resulting in a 1.3 output it looks like the current code cannot support 1.4 due to some matching case logic.

https://github.com/CycloneDX/cyclonedx-gradle-plugin/blob/e0d09d8dbb8f213563bef6b0dbb194ef218d5949/src/main/java/org/cyclonedx/gradle/CycloneDxTask.java#L488-L498

And the default still sitting at 1.3:

https://github.com/CycloneDX/cyclonedx-gradle-plugin/blob/e0d09d8dbb8f213563bef6b0dbb194ef218d5949/src/main/java/org/cyclonedx/gradle/CycloneDxTask.java#L105

### Current Behaviour

1.6.0 of this plugin implies that it defaults and supports CycloneDX schema version 1.4 via the readme, however neither the default nor explicitly setting it to 1.4 results in an 1.4 schema output.

### Expected Behaviour

Current plugin version to support schema version 1.4 and defaulting to it.

k3rnelpan1c-dev:
### Description

This PR aims to quick fix the default CycloneDx schema version to what is implied by the readme and also adds support to it properly by updating the version switch function.

### Changes

* fix function to determine schema version when set (now supports & defaults to version 1.4)
* set 1.4 to be the default CycloneDx schema
* minor nit fix for `build.gradle.kts` readme section

### Issues

* fix #154

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,8 +60,8 @@ If you are using the Kotlin DSL, the plugin can be configured as following:
 
 ```kotlin
 tasks.cyclonedxBom {
-    setIncludeConfigs(["runtimeClasspath"])
-    setSkipConfigs(["compileClasspath", "testCompileClasspath"])
+    setIncludeConfigs(listOf("runtimeClasspath"))
+    setSkipConfigs(listOf("compileClasspath", "testCompileClasspath"))
     setProjectType("application")
     setSchemaVersion("1.4")
     setDestination(project.file("build/reports"))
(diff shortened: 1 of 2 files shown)


glefloch (approved):
Thanks for this @k3rnelpan1c-dev

k3rnelpan1c-dev:
sorry had the switch-case syntax scuffed there

PR #155, opened 2022-05-28
ci: slight overhaul proposal

k3rnelpan1c-dev:
### Description

This PR aims to slightly update the existing CI and adding a workflow to keep the Gradle wrapper up to date.

I want to use this opportune to propose the idea of using Mend's Renovate Bot over GitHub Dependabot to keep dependencies and ultimately GitHub actions up to date.
The reason behind that, besides allowing for infinitely more customization, is the supports to pin GitHub Action versions and thereby improving supply chain security.

> NOTE: In case this is an idea the CycloneDX team would be open for, as the GH App needs to be added on the Org lvl first, then I would recommend using it as the global solution with presets in the `.github` repo and each project pulling from that and customizing in their repo.

### Changes

* add GradleW update workflow
* alter Build CI to build against 3 Java LTS for validation (+ add caching)
* slightly reformat the Publish CI and ensure GradleW gets validated there too
* update GradleW from `5.6.4` to `7.4.2` (to support JDK 17)
  * required a bump of the `org.spockframework:spock-core` from `1.1-groovy-2.4` to `2.2-M1-groovy-3.0`
* add [editorconfig](https://editorconfig.org)

### Issues

* n/a


glefloch:
Thanks @k3rnelpan1c-dev, everything looks good to me. Regarding the, we don't have that much dependencies and everyone knows dependabot. If we face issues we can definitely reconsider it. WDYT?

k3rnelpan1c-dev:
well, I only suggested it for features such as dependency pinning (especially in regards to GitHub Actions [and using commit hashes instead of tags](https://docs.renovatebot.com/modules/manager/github-actions/#additional-information)), but I feel indifferent overall.

I use bove over various repositories and both work great, was just an idea :wink:

PR #161, opened 2022-05-31
ci: minor cleanups and missed configs

k3rnelpan1c-dev:
### Description

This PR aims to cleam up a few loose ends left behind in #155 and specifically enables Dependabot for GitHub Actions too.

### Changes

* add GitHub Actions to Dependabot config
* ignore Groovy 4.0 dependency updated (as they are incompatible right now, might want to set that dependency to manual all together :thinking: )
* trigger build only on default-branch pushes and PRs targeting it
* migrate deprecated config values in `build.gradle` to new variant of setting them 

### Issues

* followup to #155

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -4,3 +4,12 @@ updates:
     directory: /
     schedule:
       interval: daily
+    ignore:
+      - dependency-name: org.spockframework:spock-core
+        versions:
+          # groovy 4.0 is not yet supported by Gradle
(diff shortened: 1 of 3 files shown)


k3rnelpan1c-dev:
Note: I also tried to get get the build pipeline to generate an SBOM, since the build plugin is also part of the build.gradle plugins.
However, all I got was errors that the task as not registered, so I skipped that for now.

PR #174, opened 2022-06-22
fix: javadoc warning about version on writeBom

gitphill:
Fixing the following javadoc warning

```
> Task :javadoc
cyclonedx-gradle-plugin/src/main/java/org/cyclonedx/gradle/CycloneDxTask.java:457: warning: no @param for version
    protected void writeBom(Metadata metadata, Set<Component> components, CycloneDxSchema.Version version) throws GradleException{
                   ^
1 warning
```


glefloch (approved):
Nice catch @gitphill, thanks !

Issue #203, opened 2022-08-31
StackOverflowError in CycloneDxTask#buildDependencyGraph 


S-Callier:
There is a StackOverflowError occuring in the version 1.7.0 of the plugin (does not occur with 1.6.1)

<pre>
Cause: java.lang.StackOverflowError
	at com.github.packageurl.PackageURL.validatePath(PackageURL.java:355)
	at com.github.packageurl.PackageURL.validateNamespace(PackageURL.java:260)
	at com.github.packageurl.PackageURL.validateNamespace(PackageURL.java:253)
	at com.github.packageurl.PackageURL.<init>(PackageURL.java:102)
	at org.cyclonedx.gradle.CycloneDxTask.generatePackageUrl(CycloneDxTask.java:463)
	at org.cyclonedx.gradle.CycloneDxTask.generatePackageUrl(CycloneDxTask.java:443)
	at org.cyclonedx.gradle.CycloneDxTask.buildDependencyGraph(CycloneDxTask.java:282)
	at org.cyclonedx.gradle.CycloneDxTask.buildDependencyGraph(CycloneDxTask.java:287)
	at org.cyclonedx.gradle.CycloneDxTask.buildDependencyGraph(CycloneDxTask.java:287)
	at org.cyclonedx.gradle.CycloneDxTask.buildDependencyGraph(CycloneDxTask.java:287)
	at org.cyclonedx.gradle.CycloneDxTask.buildDependencyGraph(CycloneDxTask.java:287)
	at org.cyclonedx.gradle.CycloneDxTask.buildDependencyGraph(CycloneDxTask.java:287)
	at org.cyclonedx.gradle.CycloneDxTask.buildDependencyGraph(CycloneDxTask.java:287)
</pre>

The simplest way I found to trigger the  issue is to run the `gradle cyclonedxBom` command from a project depending on `org.apache.xmlgraphics:batik-svg-dom:1.7`

I took a quick glance at the code and I assume the issue is due to a loop in the dependencies:

<pre>
+--- org.apache.xmlgraphics:batik-svg-dom:1.7
|    +--- org.apache.xmlgraphics:batik-anim:1.7
|    |    +--- org.apache.xmlgraphics:batik-awt-util:1.7
|    |    |    \--- org.apache.xmlgraphics:batik-util:1.7
|    |    +--- org.apache.xmlgraphics:batik-dom:1.7
[...]

PR #204, opened 2022-08-31
Prevent stack overflow in case of loop in the dependency graph

S-Callier:
Pull request for https://github.com/CycloneDX/cyclonedx-gradle-plugin/issues/203


glefloch (approved):
Thanks @S-Callier for this work, this looks good to me.

Issue #77, opened 2020-08-21
How to configure skipping sub-projects?


Jaff:
We have a gradle android project that pulls in a dependant sub-project, which we are scanning by default. That behavior is causing our overall scan to fail. Is there some setting that will prevent scanning the sub-project? I am using 
```
plugins {
   id 'org.cyclonedx.bom' version '1.2.0'
}
```
Build output contains:
```
BOM inclusion for configuration debugAndroidTestApiDependenciesMetadata : []
BOM inclusion for configuration debugAndroidTestCompile : []
:app:cyclonedxBom took 103ms
Task :app:cyclonedxBom in app Finished
:app:cyclonedxBom (Thread[Execution worker for ':',5,main]) completed. Took 1.782 secs.

> Task :domain:kaptGenerateStubsTestKotlin
Task :domain:kaptGenerateStubsTestKotlin in app Starting
Deleting stale output file: /sandbox/workspace/proton-android_sonarqube/fission/domain/build/tmp/kapt3/stubs/test
Build cache key for task ':domain:kaptGenerateStubsTestKotlin' is d0234ce42e26500c3061002441e4cb05
Task ':domain:kaptGenerateStubsTestKotlin' is not up-to-date because:
  No history is available.
The input changes require a full rebuild for incremental task ':domain:kaptGenerateStubsTestKotlin'.
file or directory '/sandbox/workspace/proton-android_sonarqube/fission/domain/src/test/java', not found
file or directory '/sandbox/workspace/proton-android_sonarqube/fission/domain/src/test/java', not found
Using Kotlin/JVM incremental compilation
i: found daemon on port 17962 (278000 ms old), trying to connect
i: connected to the daemon
[...]

PR #236, opened 2022-11-17
Add skipProject option

adam-siklosi:
#77 

As far as I see currently you have two options to use the plugin:

1.) Apply it on the root project
--> Cannot skip subprojects
2.) Apply it on subprojects separately
--> Produces a bom file per subproject. You need to combine those files manually later using cyclonedx-cli merge -> so why use the plugin at all
--> Combining subprojects bom results cause duplications and huge files

This PR adds an option to skip the root and subprojects.


glefloch (approved):
Thanks for this @adam-siklosi.

Issue #361, opened 2023-11-05
Version 1.8.0 doesn't actually produce SBOMs with 1.5 schema version


ThomasVitale:
I upgraded the plugin from 1.7.4 to 1.8.0 and tried to generate an SBOM following the CycloneDX 1.5 schema version, but I keep getting a 1.4 version instead (beside the wrong plugin version number as reported in https://github.com/CycloneDX/cyclonedx-gradle-plugin/issues/360).

I have tried both the default configuration (that should default to 1.5) and also setting it explicitly, but with no visible effect.

This is the configuration I used:

```groovy
plugins {
        ...
	id 'org.cyclonedx.bom' version '1.8.0'
}

tasks.cyclonedxBom {
	projectType = "application"
	outputFormat = "json"
	schemaVersion = "1.5"
}
```

The result is as follows:

```json
{
  "bomFormat" : "CycloneDX",
  "specVersion" : "1.4",
  "serialNumber" : "...",
  "version" : 1,
  "metadata" : {
    "timestamp" : "...",
    "tools" : [
      {
        "vendor" : "CycloneDX",
        "name" : "cyclonedx-gradle-plugin",
        "version" : "1.7.4"
      }
    ],
...
```

ThomasVitale:
I opened a PR to fix this issue: https://github.com/CycloneDX/cyclonedx-gradle-plugin/pull/362

PR #362, opened 2023-11-05
Fix support for CycloneDX 1.5

ThomasVitale:
Fix bug causing the CycloneDX 1.5 schema version
to be ignored by the plugin with fallback to 1.4.

Update docs with spec references to version 1.5.

Fixes gh-361

### DIFF
--- a/src/main/java/org/cyclonedx/gradle/utils/CycloneDxUtils.java
+++ b/src/main/java/org/cyclonedx/gradle/utils/CycloneDxUtils.java
@@ -4,6 +4,8 @@ import org.cyclonedx.CycloneDxSchema;
 
 public class CycloneDxUtils {
 
+    public static final CycloneDxSchema.Version DEFAULT_SCHEMA_VERSION = CycloneDxSchema.Version.VERSION_15;
+
     /**
      * Resolves the CycloneDX schema the mojo has been requested to use.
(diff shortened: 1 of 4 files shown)


src/main/java/org/cyclonedx/gradle/utils/CycloneDxUtils.java
```diff
@@ -4,6 +4,8 @@
 
 public class CycloneDxUtils {
 
+    public static final CycloneDxSchema.Version DEFAULT_SCHEMA_VERSION = CycloneDxSchema.Version.VERSION_15;
```
ThomasVitale, line 7: Centralising the default version here should reduce the likelihood to change the default version in the Task property without also updating this util class, preventing the current bug to happen again.

It's also possible to configure the default directly to `CycloneDxSchema.VERSION_LATEST`, but it would be less obvious which version is that just by reading the code, so I opted for a more explicit configuration. Thoughts?

jkowalleck (approved):
LGTM

ThomasVitale:
It looks like I missed a couple of tests to update to check against the default version rather than hard-coded 1.4. I'm about to push the missing change.

glefloch (approved):

glefloch:
Thanks @ThomasVitale!

Issue #360, opened 2023-11-04
Plugin version 1.8.0 referencing version 1.7.4 in output


0BuRner:
Seems you forgot to update the https://github.com/CycloneDX/cyclonedx-gradle-plugin/blob/master/src/main/resources/plugin.properties when releasing 1.8.0 with the corresponding version.

ThomasVitale:
I can see that when moving from Groovy to Kotlin for the Gradle configuration, the code updating the `plugin.properties` file automatically has been removed: https://github.com/CycloneDX/cyclonedx-gradle-plugin/pull/305/files#diff-49a96e7eea8a94af862798a45174e6ac43eb4f8b4bd40759b5da63ba31ec3ef7L38

It would be nice to have it automatically updated. An option is to add the task back using the Kotlin syntax and make it part of the compilation step, but it would cause continuous changes to the properties file even if nothing changed (due to the autogenerated timestamp) resulting in a poor development experience. I have drafted an alternative solution to change the file only if there was any actual change: https://github.com/CycloneDX/cyclonedx-gradle-plugin/pull/363. Any thoughts on the solution?

PR #363, opened 2023-11-05
Update plugin version used in final BOM

ThomasVitale:
Add an automated Gradle task to update the plugin.properties file as part of the compilation step, only if there was any change to version, organization, or project name.

Fixes gh-360


build.gradle.kts
```diff
@@ -37,6 +40,38 @@ tasks.withType<Test> {
 
 tasks.withType<JavaCompile>().configureEach {
     options.encoding = "UTF-8"
+    dependsOn.add(processPluginPropertiesFile)
+}
+
+val processPluginPropertiesFile = tasks.register<DefaultTask>("processPluginPropertiesFile") {
```
ThomasVitale, line 46: I can see that when moving from Groovy to Kotlin for the Gradle configuration, the code updating the plugin.properties file automatically has been removed: https://github.com/CycloneDX/cyclonedx-gradle-plugin/pull/305/files#diff-49a96e7eea8a94af862798a45174e6ac43eb4f8b4bd40759b5da63ba31ec3ef7L38

It would be nice to have it automatically updated. An option is to add the task back using the Kotlin syntax and make it part of the compilation step, but it would cause continuous changes to the properties file even if nothing changed (due to the autogenerated timestamp) resulting in a poor development experience. I have drafted this alternative solution to change the file only if there was any actual change. 

This custom task could stay here or, if we want it defined in its own file, it could be moved to a `buildSrc` folder within the project.

glefloch:
Thanks @ThomasVitale I wonder if we could just remove the `plugin.properties` file from git and simply run the task each time. Thus, we won't have to commit the new version as it will be generated for each build

ThomasVitale:
@glefloch I like that idea, much cleaner approach. I'll update the pull request.

glefloch (approved):

Issue #321, opened 2023-07-14
Unable to resolve POM for project dependencies which are not subprojects of the current project


loicrouchon:
# Context

Consider the following project structure:

* `:modules:app1`
* `:modules:app2`
* `:modules:lib1`
* `:modules:lib2`

Where both `:modules:app1` and `:modules:app2` depends on `:modules:lib1` and `:modules:lib2`.
In this case as both libs should be included in both apps, it is not possible/nor desirable to make the libs subprojects of the apps.

# The Problem

Currently the cyclonedxBom task of the cyclonedx-gradle-plugin has a shortcut for subprojects not for their POM metadata to be fetched from an artifact repository as follow:

```java
resolvedConfiguration.getResolvedArtifacts().forEach(artifact -> {
    // Don't include other resources built from this Gradle project.
    final String dependencyName = DependencyUtils.getDependencyName(artifact);
    if (builtDependencies.contains(dependencyName)) {
        return;
    }
    // fetch the dependency from a repository here
}
```

and where `builtDependencies` is defined like below:

```java
final Set<String> builtDependencies = getProject()
    .getSubprojects()
    .stream()
    .map(p -> p.getGroup() + ":" + p.getName() + ":" + p.getVersion())
    .collect(Collectors.toSet());
```

This is problematic as it means that in case the local project dependencies are not sub projects of the project running the cyclonedxBom task, CycloneDX will try to fetch them from an artifact repository.
But as they haven't been published yet (part of the same Gradle build), they will never be found.

You can see that below
```
> Task :modules:app1:cyclonedxBom
CycloneDX: Parameters
------------------------------------------------------------------------
schemaVersion          : 1.4
includeLicenseText     : true
includeBomSerialNumber : true
includeConfigs         : [runtimeClasspath]
skipConfigs            : []
skipProjects           : []
destination            : <PROJECT_DIR>/build/reports/cyclonedx
outputName             : bom
------------------------------------------------------------------------
CycloneDX: Resolving Dependencies

Resource missing. [HTTP GET: https://somenexusinstance/repository/maven-group/<group/id>/lib1/1.0.0/lib1-1.0.0.pom]
Unable to resolve POM for <group.id>:lib1:1.0.0: org.gradle.api.internal.artifacts.ivyservice.DefaultLenientConfiguration$ArtifactResolveException: Could not resolve all files for configuration ':modules:app1:detachedConfiguration1'.
[...]

PR #368, opened 2023-11-21
Add local project dependencies to the built dependencies for local resolution #321

loicrouchon:
This PR attempts to solve https://github.com/CycloneDX/cyclonedx-gradle-plugin/issues/321

This code is adding all dependencies, direct and transitive ones, which are ProjectDependency (i.e. gradle projects part of the same build) for each configurations to the builtDependencies set which otherwise only contains subprojects.


glefloch:
Thanks for this @loicrouchon ! Look good to me

DevDavido:
Would it be possible to provide a down port of this fix for the 1.7.x version? The reason is that 1.8.x is not compatible with our project Gradle version 7.5.1: https://github.com/CycloneDX/cyclonedx-gradle-plugin/issues/374

glefloch:
Yes i will try to push this quickly

PR #376, opened 2023-12-05
fix: check for artifact existence before looking inside for pom.xml

ctran:
Add a check to verify that the artifact does exist on disk before checking for the "pom.xml" inside.

Without this check, I'm getting `NoSuchFileException` in multi-module project (bar has a dependency on foo):

```
> Task :bar:cyclonedxBom
An error occurred attempting to extract POM from artifact
java.nio.file.NoSuchFileException: /Users/cuong.tran/git/gradle/hello-project/foo/build/libs/foo-25.9e298ae1.0.0.jar
        at java.base/sun.nio.fs.UnixException.translateToIOException(UnixException.java:92)
        at java.base/sun.nio.fs.UnixException.rethrowAsIOException(UnixException.java:111)
        at java.base/sun.nio.fs.UnixException.rethrowAsIOException(UnixException.java:116)
        at java.base/sun.nio.fs.UnixFileAttributeViews$Basic.readAttributes(UnixFileAttributeViews.java:55)
        at java.base/sun.nio.fs.UnixFileSystemProvider.readAttributes(UnixFileSystemProvider.java:149)
        at java.base/java.nio.file.Files.readAttributes(Files.java:1764)
        at java.base/java.util.zip.ZipFile$Source.get(ZipFile.java:1259)
        at java.base/java.util.zip.ZipFile$CleanableResource.<init>(ZipFile.java:733)
        at java.base/java.util.zip.ZipFile$CleanableResource.get(ZipFile.java:850)
        at java.base/java.util.zip.ZipFile.<init>(ZipFile.java:248)
        at java.base/java.util.zip.ZipFile.<init>(ZipFile.java:177)
        at java.base/java.util.jar.JarFile.<init>(JarFile.java:350)
        at java.base/java.util.jar.JarFile.<init>(JarFile.java:321)
        at java.base/java.util.jar.JarFile.<init>(JarFile.java:287)
        at org.cyclonedx.gradle.MavenHelper.extractPom(MavenHelper.java:270)
        at org.cyclonedx.gradle.CycloneDxTask.convertArtifact(CycloneDxTask.java:593)
        at org.cyclonedx.gradle.CycloneDxTask.lambda$null$5(CycloneDxTask.java:389)
        at java.base/java.lang.Iterable.forEach(Iterable.java:75)
        at org.cyclonedx.gradle.CycloneDxTask.lambda$createBom$6(CycloneDxTask.java:379)
```

### DIFF
--- a/src/main/java/org/cyclonedx/gradle/MavenHelper.java
+++ b/src/main/java/org/cyclonedx/gradle/MavenHelper.java
@@ -265,7 +265,7 @@ class MavenHelper {
         if (!isDescribedArtifact(artifact)) {
             return null;
         }
-        if (artifact.getFile() != null) {
+        if (artifact.getFile() != null && artifact.getFile().exists()) {
             try {
                 final JarFile jarFile = new JarFile(artifact.getFile());
                 final ModuleVersionIdentifier mid = artifact.getModuleVersion().getId();


ChristianCiach:
Is this actually the correct approach here? Sure, this suppresses the error message, but the resulting SBOM will miss the dependency on `foo`. 

In fact, the error message can easily be workarounded. Assuming that "bar" depends on "foo":

```
./gradlew :bar:cyclonedxBom
```

This will print the exceptions that this PR suppresses, because `/foo/build/libs/foo.jar` has not been built yet. Because of this, the generated `bom.json` doesn't contain the entry for `foo`.

But we can force Gradle to build all the jars first:

```
./gradlew jar :bar:cyclonedxBom
```

This fixes the exceptions and also fixes the generated SBOM.

In other words: This PR suppresses an error that should *actually* be fixed, because it results in an incomplete SBOM. Instead of suppressing the error messages, this plugin should actually add a task-dependency from `:bar:cyclonedxBom` to `:foo:jar` automatically.

ChristianCiach:
It looks like https://github.com/CycloneDX/cyclonedx-gradle-plugin/pull/368 added the project-dependencies I've proposed in my previous comment :) It probably also made this PR obsolete, because the missing jar-files will now be built automatically, so the `FileNotFoundException` won't happen anymore.

I hope a new version including the fixes will be released, soon.

Issue #336, opened 2023-09-18
Incorrect (empty) tool entry may appear in SBOMs if project also uses a plugin which uses org.eclipse.jgit


HughG:
(I originally raised this issue at https://github.com/CycloneDX/cyclonedx-core-java/issues/325 but that's the wrong project -- sorry!)

I have successfully created SBOMs from several in-house projects (can't share the source, sorry) using versions 1.7.3 and 1.7.4 of the org.cyclonedx.bom plugin. However, in one project the tools section of the SBOM contains a single entry with no properties, like this.

```
{
  "bomFormat" : "CycloneDX",
  "specVersion" : "1.4",
  "serialNumber" : "urn:uuid:9ea71a0d-3dda-40a2-99c2-d889408d03e3",
  "version" : 1,
  "metadata" : {
    "timestamp" : "2023-09-07T20:17:24Z",
    "tools" : [
      { }
    ],
...
  }
}
```
whereas in other projects it looks like this

```
...
    "tools" : [
      {
        "vendor" : "CycloneDX",
        "name" : "cyclonedx-gradle-plugin",
        "version" : "1.7.4"
      }
    ],
...
```

I don't understand the spec well enough to know if this is invalid, strictly speaking, but it's clearly not what's intended. In particular it causes https://github.com/CycloneDX/sbom-comparator to fail to compare the files, unless I manually fix or remove the incorrect entry.

Through debugging the Gradle build process, I believe the problem is that the problem project also uses com.netflix.nebula:gradle-info-plugin:9.2.0 and org.ajoberstar.grgit:grgit-gradle:4.1.0, which indirectly pulls in JAR org.eclipse.jgit:org.eclipse.jgit:6.7.0.202309050840-r. That JAR also contains a file plugin.properties at the top level of its resources folder, which of course has a completely different set of properties.

Which plugin.properties is found first on the classpath appears to depend (in my case) on the hash which Gradle computes for the JAR because that's the leftmost difference in the paths to the JARs.

I think the solution to this is that the CycloneDX plugin should move its plugin.properties to a namespaced sub-folder within the JAR, e.g., resources/org/cyclonedx/gradle/plugin.properties (or maybe a further sub-level if you have other plugins/JARs under the org.cyclonedx.gradle namespace which might have a plugin.properties.

[...]

PR #380, opened 2023-12-28
fix: move `plugin.properties` to namespaced location

gcx-seb:
This solves #336, where `plugin.properties` contributed by other dependencies overwrite the one contained in this plugin.

In my case it was [ktlint-gradle](https://github.com/JLLeitschuh/ktlint-gradle) which also contributed a `plugin.properties` at the root of the classpath (and that won over the one contributed by this plugin).

I tested this locally, and moving it to a namespaced location fixes this problem.


gcx-seb:
Anyone? :thinking:

glefloch:
Thanks for this @gcx-seb, sorry for the delay. This looks good to me.

glefloch (approved):

PR #444, opened 2024-06-06
upgrade cyclonedx-core-java to 9.x.x

skhokhlov:


skhokhlov:
Well, from my understanding, the tests are failing because of a bug in cyclonedx-java-core. 
```
CycloneDX: Validating BOM
Unknown keyword meta:enum - you should define your own Meta Schema. If the keyword is irrelevant for validation, just use a NonValidationKeyword or if it should generate annotations AnnotationKeyword
Unknown keyword deprecated - you should define your own Meta Schema. If the keyword is irrelevant for validation, just use a NonValidationKeyword or if it should generate annotations AnnotationKeyword
```
https://github.com/CycloneDX/cyclonedx-core-java/blob/master/src/main/resources/bom-1.6.schema.json#L611

That project doesn't have any tests for schema validation

skhokhlov:
Related issue https://github.com/CycloneDX/cyclonedx-core-java/issues/409

jkowalleck (approved):
LGTM.

adding support for CDX 1.6 ✔️ 
does not change any default values, or does it?

jkowalleck:
@glefloch could you review this PR and maybe merge/release it?

skhokhlov:
> does not change any default values, or does it? if it did, this would be a breaking change ...

just default schema version, will be 1.6

VinodAnandan:
@skhokhlov  can we keep 1.5 as default, maybe we can change the default version in the 1.9.x release?

skhokhlov:
> @skhokhlov can we keep 1.5 as default, maybe we can change the default version in the ~1.9.x~ 2.x release?

Any reason for this? Previously default schema version was update with minor release 1.8.0

jkowalleck:
> > @skhokhlov can we keep 1.5 as default, maybe we can change the default version in the ~1.9.x~ 2.x release?
> 
> Any reason for this? Previously default schema version was update with minor release 1.8.0

Previous changes were non-ideal. They introduced breaking changes without when users did not expect them.
We want to enable users to use CycloneDX 1.6, but don't want to force existing inexperienced users to it. 
The idea is: By not changing the default, we would enable existing users to upgrade easily, without breaking any processes.

skhokhlov:
pushed the change with 1.5 by default

jkowalleck (approved):

skhokhlov:
Any change to make it merged? 🙃

Issue #292, opened 2023-05-06
Generate reproducible outcome


aalmiray:
I'd be great if generated SBOMs were reproducible as per https://reproducible-builds.org/

This would require that:

- generated files should have a deterministic order i.e, dependencies sorted by a given rule.
- timestamp supplied as an input. Keep current behavior if no input is given.

aalmiray:
See https://github.com/CycloneDX/cyclonedx-maven-plugin/pull/353 for reference

glefloch:
Thanks @aalmiray I will work on this!

atennapel:
Hi @glefloch  , what is the status of this feature?

glefloch:
I haven't been able to work on this yet.

jdaugherty:
The [Apache Grails](https://github.com/apache/grails-core) Project [post processes](https://github.com/apache/grails-core/blob/7.0.x/build-logic/plugins/src/main/groovy/org/apache/grails/buildsrc/SbomPlugin.groovy) the output of this gradle plugin to make the generated sbom json file reproducible (amongst updating the bom to choose a specific license if multiple are present).

From my testing, I've seen the following cause reproducibility problems: 
1. `metadata.timestamp` => needs to use a fixed value, we make the build reproducible / use the value defined by [SOURCE_DATE_EPOCH](https://reproducible-builds.org/docs/source-date-epoch/) (which we set using `$(git log -1 --pretty=%ct)` )
2. `dependencies[*].dependsOn` has an nondeterministic order - we've seen the order vary across build platforms. We sort this field to fix this.
3. `dependencies[*].hashes` may not be populated in a multiproject build because the build can execute in parallel and the jar may not be generated yet for a dependency that is a project. We decided to remove project based dependencies from our sbom to make it reproducible.
4. the serial number is not reproducible. We decided to set the serial number to empty, generate a UUID from contents of the json bom file, and then set the serial number to that UUID.  So if the file is reproducible, then serial number will be as well.

PR #457, opened 2024-06-19
Order components and dependencies by purl and ref to have reproducible output

emirmx:
As the components and dependencies are stored in `HashSet` and `HashMap`, when converted to lists, the order is not well defined and seems to change from time to time, especially when run on different computers.

In this PR, components and dependencies are sorted before the BOM is written to have the same order, given same dependencies.

This should fix part of this issue: https://github.com/CycloneDX/cyclonedx-gradle-plugin/issues/292

### DIFF
--- a/src/main/java/org/cyclonedx/gradle/CycloneDxTask.java
+++ b/src/main/java/org/cyclonedx/gradle/CycloneDxTask.java
@@ -389,21 +389,16 @@ public class CycloneDxTask extends DefaultTask {
 
                 resolvedConfiguration.getResolvedArtifacts().forEach(artifact -> {
                     String dependencyName = DependencyUtils.getDependencyName(artifact);
-                    if (builtDependencies.contains(dependencyName)) {
-                        // Resources built as part of this Gradle project will be added to the
-                        // components section but not the dependencies sections
-                        Component component = convertArtifact(artifact, version);
-                        components.putIfAbsent(component.getBomRef(), component);
-                    } else {
-                        // Resources not built as part of this Gradle project will be added to the
-                        // components and dependencies sections. They will also be augmented with
-                        // metadata from their poms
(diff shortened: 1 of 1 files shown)


skhokhlov:
You can just use TeeSet instead of HashSet https://docs.oracle.com/javase/8/docs/api/java/util/TreeSet.html

emirmx:
> You can just use TreeSet instead of HashSet https://docs.oracle.com/javase/8/docs/api/java/util/TreeSet.html

Thanks, way better. Changed.

emirmx:
Hi. Is there a time-frame for PR's to be reviewed/merged? Also, for the releases?

skhokhlov (changes requested):
I think this https://github.com/CycloneDX/cyclonedx-gradle-plugin/blob/master/src/main/java/org/cyclonedx/gradle/CycloneDxTask.java#L398 can be removed now

emirmx:
> I think this https://github.com/CycloneDX/cyclonedx-gradle-plugin/blob/master/src/main/java/org/cyclonedx/gradle/CycloneDxTask.java#L398 can be removed now

Sorry, could not tie your comment to the changes in the PR. Can you elaborate?

emirmx:
> I think this https://github.com/CycloneDX/cyclonedx-gradle-plugin/blob/master/src/main/java/org/cyclonedx/gradle/CycloneDxTask.java#L398 can be removed now

Though I still do not see the connection, I simplified the code in createBom a bit more within two additional commits.

skhokhlov (approved):
Sorry, the code moved a little bit. I meant this https://github.com/CycloneDX/cyclonedx-gradle-plugin/pull/457/files#diff-7d5ff0e9f240ba234a7a8e596f574d6fe41631e6fb16f8a1d2a252a6ae5ac72fL413 which is removed now. Thank you!

Issue #466, opened 2024-07-22
Enable override of the default component name


MH-17:
The component name is automatically set to the project name when the SBOM is generated. In the same manner that you can override the component version, I want to allow users to alter this component name.

Background: The same component is being customized for several projects at my company. It is the goal of the developers to open several of these projects in a single Eclipse workspace. They can only do this if their Gradle project names differ. As a result, we rely on the directory name in which the project is kept rather than having a static project name declared in our repositories. Thus far, everything is going according to plan. However, our Azure DevOps pipelines check out the code into a directory named "s", which causes SBOMs to be generated with the component name "s". Above change would allow us to define a static component name for the CycloneDxTask.

I would be willing to provide a PR for this change. Please let me know, if you would accept this change. :)

skhokhlov:
Well, I believe that current behaviour to have the component name same as project name is pretty correct. It just explicitly represents gradle project metadata in SBOM.

You can configure project name on the gradle side [for root project](https://docs.gradle.org/current/userguide/writing_settings_files.html#the_settings_object) and [for any other project](https://docs.gradle.org/current/userguide/writing_build_scripts.html#sec:standard_project_properties).

Please let me know if I'm missing something.

skhokhlov:
Related to https://github.com/CycloneDX/cyclonedx-gradle-plugin/issues/446

PR #467, opened 2024-07-22
Enable override of the default component name

MH-17:
Proposal for #466


skhokhlov (changes requested):
Can you please add a test to verify that the current behaviour is still there if the property is not set? We will merge it after it.

skhokhlov (approved):
Thank you for the update!

Issue #366, opened 2023-11-09
I made an error building the BOM table for multiple projects by using the --init-script option. I don't know what happened


wzd-hash:
`CycloneDX: Resolving Dependencies
An unexpected issue occurred attempting to create a PackageURL for :smart-marketing-resource:unspecified
An unexpected issue occurred attempting to create a PackageURL for :smart-marketing-resource:unspecified
An unexpected issue occurred attempting to create a PackageURL for :smart-marketing-resource:unspecified
CycloneDX: Creating BOM
CycloneDX: Writing BOM JSON
CycloneDX: Validating BOM
Task :cyclonedxBom in app Finished
:cyclonedxBom spend 291ms
:cyclonedxBom (Thread[Execution worker for ':',5,main]) completed. Took 0.312 secs.
      00:00.29   :cyclonedxBom

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':cyclonedxBom'.
> The BOM does not conform to the CycloneDX BOM standard

* Try:
Run with --debug option to get more log output. Run with --scan to get full insights.

* Exception is:
org.gradle.api.tasks.TaskExecutionException: Execution failed for task ':cyclonedxBom'.
        at org.gradle.api.internal.tasks.execution.ExecuteActionsTaskExecuter.lambda$executeIfValid$1(ExecuteActionsTaskExecuter.java:205)
        at org.gradle.internal.Try$Failure.ifSuccessfulOrElse(Try.java:263)
        at org.gradle.api.internal.tasks.execution.ExecuteActionsTaskExecuter.executeIfValid(ExecuteActionsTaskExecuter.java:203)
        at org.gradle.api.internal.tasks.execution.ExecuteActionsTaskExecuter.execute(ExecuteActionsTaskExecuter.java:184)
        at org.gradle.api.internal.tasks.execution.CleanupStaleOutputsExecuter.execute(CleanupStaleOutputsExecuter.java:109)
        at org.gradle.api.internal.tasks.execution.FinalizePropertiesTaskExecuter.execute(FinalizePropertiesTaskExecuter.java:46)
        at org.gradle.api.internal.tasks.execution.ResolveTaskExecutionModeExecuter.execute(ResolveTaskExecutionModeExecuter.java:62)
        at org.gradle.api.internal.tasks.execution.SkipTaskWithNoActionsExecuter.execute(SkipTaskWithNoActionsExecuter.java:57)
[...]

PR #469, opened 2024-07-22
fix: fail plugin execution if group, name, or version are not set

skhokhlov:
Currently the execution is failing anyway, but without providing clear reason to the user


skhokhlov:
covers https://github.com/CycloneDX/cyclonedx-gradle-plugin/issues/366

Issue #432, opened 2024-04-30
ProjectDependency are missing from the components section of the BOM


loicrouchon:
Hello,

In https://github.com/CycloneDX/cyclonedx-gradle-plugin/pull/368, I proposed a fix for https://github.com/CycloneDX/cyclonedx-gradle-plugin/issues/321 which was merged and released.

The fix was a generalization of the sub-projects handling to all projects part of the same Gradle build (i.e. dependencies of type `ProjectDependency`).

This is done by adding such coordinates for those `ProjectDependency` to the `builtDependencies` local variable in `org.cyclonedx.gradle.CycloneDxTask#createBom`. This avoids trying to fetch the BOM for those dependencies remotely as they are local ones.

However, I realized a side effect of doing this was resulting in local project dependencies not being part of the `"components"` section of the BOM and I don't think this is the intended behavior as it means those would still be expressed in the `"dependencies"` section, but without a way to check the licenses nor the hashes of such dependencies in the `"components"` section.

Could you confirm if this is intended behavior? If not I can try to contribute a fix.

ericparton:
I've noticed this behavior and as a user, I feel that it's a bug. The fact that these dependencies don't appear in the components section causes odd behavior in third party tools that consume the boms

HoustonPutman:
Yes, I've seen this too and also think it's a bug.

loicrouchon:
Just a heads-up that I finally have some time to invest into fixing this issue. I made good progress today and I now need to write extensive tests for the prototype. I'll open a PR (hopefully next week) when done with the tests.

PR #474, opened 2024-07-29
Include local project dependencies as components #432

loicrouchon:
This PR attempts to solve #432

This is done by:

* Simplifying how all projects are added to the `builtDependencies`. Fix #368 which only added direct local project dependencies, but forgot to add transitive ones to local project.
* Not generating the `generateProjectComponent` at the top, but putting it at the end, in a second loop. This allows to compute extra information for the components (for example hashes) if it is part of the dependency graph.
* Ensure dependencies part of `builtDependencies` are added to the `components` field of the BOM, but without triggering any maven lookup for metadata.


skhokhlov (approved):
LGTM. Thank you @loicrouchon

Issue #416, opened 2024-04-08
Should default config include test scope? 


blommish:
When generating bom with default config, the test scope is included. 
https://github.com/CycloneDX/cyclonedx-gradle-plugin/compare/master...blommish:cyclonedx-gradle-plugin:default-config-including-test-scope?expand=1

In the maven-plugin, the test scope is not included by default
https://github.com/CycloneDX/cyclonedx-maven-plugin?tab=readme-ov-file#default-values

glefloch:
I will remove it to match the maven plugin behaviour

blommish:
Any updates on this?

VinodAnandan:
@skhokhlov  labeled it as a "good first issue". I think it means that anyone is welcome to contribute to the PR.

PR #505, opened 2024-09-06
Default config should not include test classpath

blommish:
Connected to 
* https://github.com/CycloneDX/cyclonedx-gradle-plugin/issues/416

https://github.com/CycloneDX/cyclonedx-maven-plugin does not include the test scope as default config. 

### What has been done
* To allow setting the list to emptylist, it makes most sense to let `set` methods overwrite previous values and not use `addAll`
* Adding `"runtimeClasspath", "compileClasspath", "annotationProcessor"` as default configs. 
  * These will also include compileOnly, runtimeOnly

### For discussion:
Would the classpaths I've added be enough? Or should others also be added? Buildscript, plugin and sourceSet.
I'm not totally sure just having the 3 I've added is enough. 
Or would it be better to just add the test-paths to ignore? 

I'll wait to update the readme until I know if this could be used or not.


VinodAnandan:
Hi @blommish  Thank you for your PR. Could you please address the DCO failure?
https://github.com/CycloneDX/cyclonedx-gradle-plugin/pull/505/checks?check_run_id=29790906669

src/main/java/org/cyclonedx/gradle/CycloneDxTask.java
```diff
@@ -135,6 +134,7 @@ public CycloneDxTask() {
         componentVersion.convention(getProject().getVersion().toString());
 
         includeConfigs = getProject().getObjects().listProperty(String.class);
+        includeConfigs.addAll("runtimeClasspath", "compileClasspath", "annotationProcessor");
```
skhokhlov, line 137: I think it should all configurations, but test ones. Otherwise it skips not only tests.

skhokhlov (changes requested):
Hi @blommish Thank you for your PR. There is problem here that this is quite a breaking change of the plugin behaviour. So I believe the best option would be to wait until the next major version release.

src/main/java/org/cyclonedx/gradle/CycloneDxTask.java
blommish: Not sure I understand what you mean. 

I agree that it is an breaking change.
The issue was that the plugin contains configs that are not included in the maven plugin. https://github.com/CycloneDX/cyclonedx-gradle-plugin/issues/416#issuecomment-2071903464

skhokhlov: The problem is that you cannot fully match maven plugin behavior just because it's a different build system. Using Gradle, you can custom configurations and use them during runtime. Using proposed changed to the default config, plugin will skip it which is incorrect.

blommish: I will close this, since nothing will happen will happen with this

PR #520, opened 2024-09-30
feat: add git external reference.

barblin:


src/main/java/org/cyclonedx/gradle/utils/ExternalResourcesUtil.java
```diff
@@ -0,0 +1,73 @@
...
+            return null;
+        }
+
+        if (gitUrl.startsWith("git@")) {
+            final String[] parts = gitUrl.split("git@");
+            String domain = parts[1];
+            return "https://" + domain.replace(":", "/");
+        }
+
+        System.out.println("REMOTE URL: " + gitUrl);
```
skhokhlov, line 70: Plugin must avoid producing unnecessary output. This message looks like a debug log

src/main/java/org/cyclonedx/gradle/utils/ExternalResourcesUtil.java
```diff
@@ -0,0 +1,73 @@
...
+        }
+    }
+
+    /**
+     * Converts a git based git remote url to a https based url. In case the url is null, map will return null
+     *
+     * @param gitUrl The git url of the current project
+     * @return gitUrl Mapped to https based git reference. May be null. Null will result in an empty external resources list
+     */
+    public static String mapGitToHttps(String gitUrl) {
```
skhokhlov, line 59: Add `@Nullable` annotation to method that can return `null`

src/main/java/org/cyclonedx/gradle/utils/ExternalResourcesUtil.java
```diff
@@ -0,0 +1,73 @@
...
+        if (gitUrl == null) {
+            return null;
+        }
+
+        if (gitUrl.startsWith("git@")) {
+            final String[] parts = gitUrl.split("git@");
+            String domain = parts[1];
+            return "https://" + domain.replace(":", "/");
```
skhokhlov, line 67: It's a dangerous assumption that `http` remote will be same as `git` one. I certainly know that it's incorrect for some product

src/main/java/org/cyclonedx/gradle/utils/ExternalResourcesUtil.java
```diff
@@ -0,0 +1,73 @@
...
+ */
+package org.cyclonedx.gradle.utils;
+
+import java.io.BufferedReader;
+import java.io.InputStreamReader;
+import java.util.ArrayList;
+import java.util.List;
+import org.cyclonedx.model.ExternalReference;
+
+public class ExternalResourcesUtil {
```
skhokhlov, line 27: This class requires benchmarking. I afraid it can significantly increase BOM generation time. You can use gradle profiler for it https://github.com/gradle/gradle-profiler

src/main/java/org/cyclonedx/gradle/utils/ExternalResourcesUtil.java
```diff
@@ -0,0 +1,73 @@
...
+
+    public static List<ExternalReference> createExternalReferences() {
+        final List<ExternalReference> references = new ArrayList<>();
+        addGitReference(references);
+        return references;
+    }
+
+    public static void addGitReference(final List<ExternalReference> references) {
+        try {
+            Process process = Runtime.getRuntime().exec("git remote get-url origin");
```
skhokhlov, line 39: Will it work for any OS?

src/main/java/org/cyclonedx/gradle/CycloneDxTask.java
```diff
@@ -53,25 +53,16 @@
 import org.cyclonedx.generators.xml.BomXmlGenerator;
 import org.cyclonedx.gradle.utils.CycloneDxUtils;
 import org.cyclonedx.gradle.utils.DependencyUtils;
-import org.cyclonedx.model.Bom;
```
skhokhlov, line 56: Please don't use "star" imports

src/test/groovy/org/cyclonedx/gradle/utils/ExternalResourcesUtilTest.groovy
```diff
@@ -0,0 +1,50 @@
...
+import org.cyclonedx.model.Bom
+import org.gradle.testkit.runner.GradleRunner
+import org.gradle.testkit.runner.TaskOutcome
+import spock.lang.Specification
+
+class ExternalResourcesUtilTest extends Specification {
+
+    def "should map git@ url to https based external reference"() {
+        expect:
+        ExternalResourcesUtil.mapGitToHttps("git@github.com:barblin/cyclonedx-gradle-plugin.git") == "https://github.com/barblin/cyclonedx-gradle-plugin.git"
```
skhokhlov, line 14: Use `assert` here

src/test/groovy/org/cyclonedx/gradle/utils/ExternalResourcesUtilTest.groovy
```diff
@@ -0,0 +1,50 @@
+package org.cyclonedx.gradle.utils
+
+import com.fasterxml.jackson.databind.ObjectMapper
+import org.cyclonedx.gradle.TestUtils
+import org.cyclonedx.model.Bom
+import org.gradle.testkit.runner.GradleRunner
+import org.gradle.testkit.runner.TaskOutcome
+import spock.lang.Specification
+
+class ExternalResourcesUtilTest extends Specification {
```
skhokhlov, line 10: The following test scenarios are required here:

1. Package without git reference
2. Package with VCS reference of different type, e.g. SVN
3. Package with incorrect git reference 
4. `git` executable not available on the system

skhokhlov (changes requested):

src/main/java/org/cyclonedx/gradle/utils/ExternalResourcesUtil.java
barblin: the conversion is actually not necessary, I have made some changes to address these issues in the PR update

src/main/java/org/cyclonedx/gradle/utils/ExternalResourcesUtil.java
barblin: Yes, should work on any OS, but I will give it a shot on my windows machine as well. I wrote some test cases to reproduce these cases

src/test/groovy/org/cyclonedx/gradle/utils/ExternalResourcesUtilTest.groovy
barblin: I added tests for most of what you've asked. Due to the fact that I am only intersted in the git VCS for now, I have ignored SVN altogether. In case a git url can't be resolved, the external reference will just not be added.

src/main/java/org/cyclonedx/gradle/utils/ExternalResourcesUtil.java
barblin: @skhokhlov 
On a large project, I have run each case and named the file accordingly. 
It seems, that adding this feature does not have a large impact on performance.
I also checked each generated sbom and each case works fine. the change can successfully pull the remote url either from custom configuration, from environment variables (e.g. Jenkins) and from the git repo config. In this order.

For completeness sake, I have also attached the scenarios file.
[performance.scenarios.txt](https://github.com/user-attachments/files/17957833/performance.scenarios.txt)

[benchmark-v1.10.0.csv](https://github.com/user-attachments/files/17957842/benchmark-v1.10.0.csv)
[benchmark-custom-configuration.csv](https://github.com/user-attachments/files/17957848/benchmark-custom-configuration.csv)
[benchmark-from-env-var.csv](https://github.com/user-attachments/files/17957851/benchmark-from-env-var.csv)
[benchmark-from-repo-config.csv](https://github.com/user-attachments/files/17957855/benchmark-from-repo-config.csv)

src/main/java/org/cyclonedx/gradle/utils/ExternalReferencesUtil.java
```diff
@@ -0,0 +1,85 @@
...
+            try (BufferedReader reader = new BufferedReader(new InputStreamReader(process.getInputStream()))) {
+                return reader.readLine();
+            }
+        } catch (Exception ignore) {
+            // throw it away if this is not a git repository or if git is not installed
+        }
+
+        return null;
+    }
+}
```
jeremylong, line 85: Minor suggestion - while `Runtime.getRuntime().exec(...` works well in most cases I have run into issues with not using a StreamGobbling technique:

```suggestion
    private static @Nullable String fromGitRepo() {
        try {
            ProcessBuilder processBuilder = new ProcessBuilder("git", "remote", "get-url", "origin");
            Process process = processBuilder.start();

            StreamGobbler outputGobbler = new StreamGobbler(process.getInputStream());
            StreamGobbler errorGobbler = new StreamGobbler(process.getErrorStream());

            outputGobbler.start();
            errorGobbler.start();

            int exitCode = process.waitFor();

            outputGobbler.join();
            errorGobbler.join();

            if (exitCode == 0) {
                return outputGobbler.getOutput().trim();
            }
        } catch (Exception ignore) {
            // throw it away if this is not a git repository or if git is not installed
        }
        return null;
    }

    private static class StreamGobbler extends Thread {
        private final InputStream inputStream;
        private final StringBuilder output = new StringBuilder();

        StreamGobbler(InputStream inputStream) {
            this.inputStream = inputStream;
        }

        @Override
        public void run() {
            try (BufferedReader reader = new BufferedReader(new InputStreamReader(inputStream))) {
                String line;
                while ((line = reader.readLine()) != null) {
                    output.append(line);
                }
            } catch (IOException e) {
                // Handle or log exception as needed
            }
        }

        String getOutput() {
            return output.toString();
        }
    }
}
```

src/main/java/org/cyclonedx/gradle/utils/ExternalReferencesUtil.java
```diff
@@ -0,0 +1,85 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ * Copyright (c) OWASP Foundation. All Rights Reserved.
+ */
+package org.cyclonedx.gradle.utils;
+
+import java.io.BufferedReader;
+import java.io.InputStreamReader;
```
jeremylong, line 22: In addition to my suggestion below, two imports would need to be added.

```suggestion
import java.io.IOException;
import java.io.InputStream;
import java.io.InputStreamReader;
```

jeremylong (commented):
Minor suggestion to use StreamGobblers when reading from a process.

src/main/java/org/cyclonedx/gradle/utils/ExternalReferencesUtil.java
barblin: Hey. I would like to know more about those issues you experienced. 

I'd rather go for simplicity if possible, and add no git reference if I run into a low probability issue. But if your issues occur often, the change may make sense.

src/main/java/org/cyclonedx/gradle/utils/ExternalReferencesUtil.java
jeremylong: This is likely overkill for this shell command. However, in other cases you can run into issues if one of the streams (error or output) fills up it can cause the process to block/hang.

src/main/java/org/cyclonedx/gradle/utils/ExternalReferencesUtil.java
jeremylong: This is likely overkill for this shell command. However, in other cases you can run into issues if one of the streams (error or output) fills up it can cause the process to block/hang. 

This is a very [common technique](https://www.google.com/search?q=java+streamgobbler+process) used to read from a process.

jeremylong:
@skhokhlov I think this PR is good as is! I'll be submitting another PR related to the build_system shortly.

jeremylong (approved):

src/main/java/org/cyclonedx/gradle/utils/ExternalReferencesUtil.java
```diff
@@ -0,0 +1,85 @@
...
+        }
+    }
+
+    private static @Nullable String fromEnvironment() {
+        return System.getenv("GIT_URL");
+    }
+
+    private static @Nullable String fromGitRepo() {
+        try {
+            Process process = Runtime.getRuntime().exec("git remote get-url origin");
```
skhokhlov, line 74: [`exec(String)` is deprecated since Java 18 ](https://cr.openjdk.org/~pminborg/panama/21/v1/javadoc/java.base/java/lang/Runtime.html#exec(java.lang.String)). It's better to use `exec(String[])` instead

src/main/java/org/cyclonedx/gradle/CycloneDxTask.java
```diff
@@ -336,6 +345,23 @@ public void setLicenseChoice(final Consumer<LicenseChoice> customizer) {
         getInputs().property("LicenseChoice", licenseChoice);
     }
 
+    public void setVCSGit(Consumer<ExternalReference> customizer) {
```
skhokhlov, line 348: Codebase in this project uses `final` keywords whenever possible. There are few places here where `final` is missing. Unfortunately we don't have automated check here, because with Java 8 it's little bit tricky nowadays.

skhokhlov (changes requested):
Thank you @barblin. There are just few minor notes left and I will be happy to merge it!

skhokhlov (approved):

PR #527, opened 2024-10-21
Disable dependency verification during each task when resolving as part of SBOM generation.

kennethshackleton:
Gradle's dependency verification can have already happened elsewhere by the time these tasks run and its not really the responsibility of the report tasks to take on any of this work.

The `disableDependencyVerification` method was added in Gradle 6.2:
https://docs.gradle.org/current/kotlin-dsl/gradle/org.gradle.api.artifacts/-resolution-strategy/disable-dependency-verification.html

xref https://docs.gradle.org/current/userguide/dependency_verification.html


skhokhlov:
Disabling verification of dependencies means that the results SBOM is vulnerable for supply chain attack, because dependency metadata can be modified. So considering potential risks we should have clear benefits of disabling verification.

PR #544, opened 2024-12-02
fix: avoid using deprecated tools section

jeremylong:
Instead of using the tools section deprecated in 1.5:

```json
"metadata" : {
    "timestamp" : "2024-12-02T22:12:31Z",
    "tools" : [
      {
        "vendor" : "CycloneDX",
        "name" : "cyclonedx-gradle-plugin",
        "version" : "1.10.0"
      }
    ],
```

If using schema 1.5+ the non-deprecated version of tools is used:

```json
  "metadata" : {
    "timestamp" : "2024-12-02T22:15:04Z",
    "tools" : {
      "components" : [
        {
          "type" : "application",
          "author" : "CycloneDX",
          "name" : "cyclonedx-gradle-plugin",
          "version" : "1.11.1"
        }
      ]
    },
```

Note that this PR also begins the work to parallelize the test cases; to improve the build performance the test cases in `DependencyResolutionSpec.groovy` and `PluginConfigurationSpec.groovy` should be moved to their own `*Spec.groovy` file so that they can be run in parallel.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -38,6 +38,7 @@ dependencies {
 
 tasks.withType<Test> {
     useJUnitPlatform()
+    maxParallelForks = (Runtime.getRuntime().availableProcessors() / 2).coerceAtLeast(1)
 }
 
 tasks.withType<JavaCompile>().configureEach {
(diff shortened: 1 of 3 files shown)


skhokhlov (changes requested):
I'm pretty happy with these changes. However, just one thing. Can you please add a test that old tools section is used for schema versions below 1.5?

jeremylong:
Thank you for the suggested test case - found a [bug](https://github.com/CycloneDX/cyclonedx-gradle-plugin/pull/544/commits/a80edd27886e1446d3b3e6bc9be2d61a7bf87fbf) that affects not only this PR but also how the hashes are calculated (see `SbomBuilder.calculateHashes()`).

skhokhlov (approved):

PR #546, opened 2024-12-05
feat: retrieve build-system metadata from build environment

jeremylong:
Allows the gradle plugin to retrieve the build-system URL from the environment variables present in most CI systems. By default, this feature is disabled. However, for those who would like to include this information in the generated bom they can configure:

```groovy
cyclonedxBom {
    includeBuildSystem = true
}
```

This will attempt to retrieve the build job URL from the CI's environment variables for several CI systems (Jenkins, GitHub Actions, Drone, Jenkins, Circle CI, and Travis CI). If not found, the `build-system` meta-data is not populated. If you are using an different CI system not listed, you can configure the feature to use alternative environment variables and even a pattern of multiple environment variables:

```groovy
cyclonedxBom {
    includeBuildSystem = true
    buildSystemEnvironmentVariable = 'MY_CUSTOM_CI_URL'
}
```

Using multiple environment variables can be achieved using the following (note that you must use single quotes to prevent gradle from interpolating the variables):

```groovy
cyclonedxBom {
    includeBuildSystem = true
    buildSystemEnvironmentVariable = '${SERVER}/build/${BUILD_ID}'
}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,6 +65,10 @@ cyclonedxBom {
     includeLicenseText = false
     // Include resolution of full metadata for components including licenses. Defaults to 'true'
     includeMetadataResolution = true
+    // Attempt to include the build-system URL by reading environment variables from common CI system such as GitHub Actions, GitLab CI, Drone, Jenkins, Travis CI, and Circle CI. Defaults to 'false'
+    includeBuildSystem = true
+    // if includeBuildSystem is true, the given environment variables will be used to construct the build-system URL that will be included in the BOM. Optional, defaults to `null`.
+    buildSystemEnvironmentVariable = '${CUSTOM_CI_URL}/jobs/${CUSTOM_JOB_ID}'
     // Override component version. Defaults to the project version
     componentVersion = "2.0.0"
     // Override component name. Defaults to the project name
--- a/README.md
+++ b/README.md
@@ -67,7 +67,7 @@ cyclonedxBom {
     includeMetadataResolution = true
     // Attempt to include the build-system URL by reading environment variables from common CI system such as GitHub Actions, GitLab CI, Drone, Jenkins, Travis CI, and Circle CI. Defaults to 'false'
     includeBuildSystem = true
-    // if includeBuildSystem is true, the given environment variables will be used to construct the build-system URL that will be included in the BOM. Optional, defaults to `null`.
+    // if includeBuildSystem is true, the given environment variables will be used to construct the build-system URL that will be included in the BOM. The dollar sign and curly braces (e.g. `${NAME}`) are required to specify an environment variable. Optional, defaults to `null`.
     buildSystemEnvironmentVariable = '${CUSTOM_CI_URL}/jobs/${CUSTOM_JOB_ID}'
     // Override component version. Defaults to the project version
     componentVersion = "2.0.0"
--- a/src/main/java/org/cyclonedx/gradle/SbomBuilder.java
+++ b/src/main/java/org/cyclonedx/gradle/SbomBuilder.java
@@ -153,7 +153,7 @@ class SbomBuilder {
         return metadata;
     }
 
-    private void addBuildSystemMetaData(Component component) {
+    private void addBuildSystemMetaData(final Component component) {
         if (task.getIncludeBuildSystem().get()) {
             String url;
             if (task.getBuildSystemEnvironmentVariable().isPresent()
--- a/src/main/java/org/cyclonedx/gradle/SbomBuilder.java
+++ b/src/main/java/org/cyclonedx/gradle/SbomBuilder.java
@@ -167,10 +167,7 @@ class SbomBuilder {
                 ExternalReference buildRef = new ExternalReference();
                 buildRef.setType(ExternalReference.Type.BUILD_SYSTEM);
                 buildRef.setUrl(url);
-                if (component.getExternalReferences() == null) {
-                    component.setExternalReferences(new ArrayList<>());
-                }
-                component.getExternalReferences().add(buildRef);
+                component.addExternalReference(buildRef);
             }
         }
     }
--- a/src/main/java/org/cyclonedx/gradle/SbomBuilder.java
+++ b/src/main/java/org/cyclonedx/gradle/SbomBuilder.java
@@ -41,6 +41,7 @@ import org.cyclonedx.gradle.model.SbomComponent;
 import org.cyclonedx.gradle.model.SbomComponentId;
 import org.cyclonedx.gradle.utils.CycloneDxUtils;
 import org.cyclonedx.gradle.utils.DependencyUtils;
+import org.cyclonedx.gradle.utils.EnvironmentUtils;
 import org.cyclonedx.model.Bom;
 import org.cyclonedx.model.Component;
 import org.cyclonedx.model.Dependency;
@@ -114,6 +115,7 @@ class SbomBuilder {
             component.setProperties(null);
             component.setName(task.getComponentName().get());
             component.setVersion(task.getComponentVersion().get());
+            addBuildSystemMetaData(component);
             metadata.setComponent(component);
         } catch (MalformedPackageURLException e) {
             logger.warn(
@@ -146,6 +148,28 @@ class SbomBuilder {
         }
 
         return metadata;
+    }
+
+    private void addBuildSystemMetaData(Component component) {
+        if (task.getIncludeBuildSystem().get()) {
+            String url;
+            if (task.getBuildSystemEnvironmentVariable().isPresent()
+                    && task.getBuildSystemEnvironmentVariable().get() != null) {
+                url = EnvironmentUtils.getBuildURI(
+                        task.getBuildSystemEnvironmentVariable().get());
+            } else {
+                url = EnvironmentUtils.getBuildURI();
+            }
+            if (url != null) {
+                ExternalReference buildRef = new ExternalReference();
+                buildRef.setType(ExternalReference.Type.BUILD_SYSTEM);
+                buildRef.setUrl(url);
+                if (component.getExternalReferences() == null) {
+                    component.setExternalReferences(new ArrayList<>());
+                }
+                component.getExternalReferences().add(buildRef);
+            }
+        }
     }
 
     private void addDependency(final Set<Dependency> dependencies, final SbomComponent component) {
--- a/src/main/java/org/cyclonedx/gradle/utils/EnvironmentUtils.java
+++ b/src/main/java/org/cyclonedx/gradle/utils/EnvironmentUtils.java
@@ -1,0 +1,129 @@
+/*
+ * This file is part of CycloneDX Gradle Plugin.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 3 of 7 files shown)


jeremylong:
I have not updated the `README.md` yet - if this is merged before a release, the documentation will be for the next release. Let me know if you would like me to update the documentation in this PR.

jeremylong:
I did not make this a configurable field, as in my use cases we want to be able to track back to the actual build that produced the SBOM.

src/main/java/org/cyclonedx/gradle/SbomBuilder.java
```diff
@@ -151,6 +153,25 @@ private Metadata buildMetadata(final SbomComponent parentComponent) {
         return metadata;
     }
 
+    private void addBuildSystemMetaData(Component component) {
```
barblin, line 156: I think its generally advised to use final in this project wherever possible

src/main/java/org/cyclonedx/gradle/utils/EnvironmentUtils.java
```diff
@@ -0,0 +1,129 @@
...
+
+    /**
+     * Get the URI of the current build from the specified environment variable.
+     * Additionally, a pattern can be provided to extract the URI from the environment variable.
+     * An example pattern could be "${SERVER}/jobs/${JOB_ID}".
+     *
+     * @param str the name of the environment variable or pattern to use
+     * @return the URI of the current build or null if it cannot be determined
+     */
+    public static String getBuildURI(String str) {
```
barblin, line 123: This method is called before `getBuildUriFromPattern` and is `public`, I think you should define this method before `private static String getBuildUriFromPattern(String pattern) {`.

Also `isNullOrBlank` does not contribute to the public interface and is only a helper method, it could be listed last in the file

src/main/java/org/cyclonedx/gradle/utils/EnvironmentUtils.java
```diff
@@ -0,0 +1,129 @@
...
+    private static boolean isNullOrBlank(String value) {
+        return value == null || value.trim().isEmpty();
+    }
+
+    /**
+     * Get the URI of the current build from the environment variables set on common build systems like GitHub Actions, GitLab CI, etc.
+     * @return the URI of the current build or null if it cannot be determined
+     */
+    public static String getBuildURI() {
+        // GitHub Actions
```
barblin, line 40: if you extract each code block into a method, you could name the method in a similar fashion as your comment, and drop the comment. for example

```
        return Optional.ofNullable(fromGithubActions())
                       .orElseGet(() -> Optional.ofNullable(fromGitlabCI()))
                       .orElseGet(() -> Optional.ofNullable(fromJenkins()))
                         ...
                       .orElse(null); 
```

```
private static final String fromGithubActions() {
        String githubServerUrl = System.getenv("GITHUB_SERVER_URL");
        String githubRepository = System.getenv("GITHUB_REPOSITORY");
        String githubRunId = System.getenv("GITHUB_RUN_ID");

        if (!isNullOrBlank(githubServerUrl) && !isNullOrBlank(githubRepository) && !isNullOrBlank(githubRunId)) {
            return String.format("%s/%s/actions/runs/%s", githubServerUrl, githubRepository, githubRunId);
        }
        
        return null;
}

```

src/main/java/org/cyclonedx/gradle/utils/EnvironmentUtils.java
```diff
@@ -0,0 +1,129 @@
...
+        if (!isNullOrBlank(ciProjectUrl) && !isNullOrBlank(ciJobId)) {
+            return String.format("%s/-/jobs/%s", ciProjectUrl, ciJobId);
+        }
+
+        // Jenkins
+        String buildUrl = System.getenv("BUILD_URL");
+        if (!isNullOrBlank(buildUrl)) {
+            return buildUrl;
+        }
+        //        String jobUrl = System.getenv("JOB_URL");
```
barblin, line 61: did you leave this comments by mistake?

src/main/java/org/cyclonedx/gradle/utils/EnvironmentUtils.java
```diff
@@ -0,0 +1,129 @@
...
+    /**
+     * Get the URI of the current build from the specified environment variable.
+     * Additionally, a pattern can be provided to extract the URI from the environment variable.
+     * An example pattern could be "${SERVER}/jobs/${JOB_ID}".
+     *
+     * @param str the name of the environment variable or pattern to use
+     * @return the URI of the current build or null if it cannot be determined
+     */
+    public static String getBuildURI(String str) {
+        if (str.contains("${")) {
```
barblin, line 124: str could be null, that could raise a NullPointerException

barblin: we could consider a test case for str being null or empty

src/main/java/org/cyclonedx/gradle/utils/EnvironmentUtils.java
```diff
@@ -0,0 +1,129 @@
...
+
+    /**
+     * Get the URI of the current build from the specified pattern and environment variables and pattern. An example
+     * pattern could be "${SERVER}/jobs/${JOB_ID}". The environment variables are specified as a list of names.
+     * @param pattern the pattern to use to build the URI
+     * @return the URI of the current build or null if it cannot be determined
+     */
+    private static String getBuildUriFromPattern(String pattern) {
+        String result = pattern;
+        Pattern regex = Pattern.compile("\\$\\{([^}]+)\\}");
```
barblin, line 98: Pattern could be a constant (as well as matcher)

src/main/java/org/cyclonedx/gradle/SbomBuilder.java
```diff
@@ -151,6 +153,25 @@ private Metadata buildMetadata(final SbomComponent parentComponent) {
...
+        if (task.getIncludeBuildSystem().get()) {
+            String url;
+            if (task.getBuildSystemEnvironmentVariable().isPresent()
+                    && task.getBuildSystemEnvironmentVariable().get() != null) {
+                url = EnvironmentUtils.getBuildURI(
+                        task.getBuildSystemEnvironmentVariable().get());
+            } else {
+                url = EnvironmentUtils.getBuildURI();
+            }
+            if (url != null) {
```
barblin, line 166: The addExternalReference method would not add the external reference if the url is not a valid Url. I think it makes no difference if it is null, or just an invalid url. 

In case I am wrong, then we should also check if the url is blank imo

barblin (changes requested):
great :+1: , just left some comments.

src/main/java/org/cyclonedx/gradle/utils/EnvironmentUtils.java
jeremylong: Thanks, I'll remove the `IsNullOrBlank()` - I missed that commons lang3 is on the classpath.

src/main/java/org/cyclonedx/gradle/utils/EnvironmentUtils.java
jeremylong: Thanks for the feedback! The code has been updated and is a little more elegant now ;)

src/main/java/org/cyclonedx/gradle/SbomBuilder.java
jeremylong: Quick glance through the code - I think if the url is blank it will likely still add it to the model.

barblin (approved):
looks good to me

src/main/java/org/cyclonedx/gradle/utils/EnvironmentUtils.java
```diff
@@ -0,0 +1,146 @@
...
+     * Get the URI of the current build from the environment variables set on common build systems like GitHub Actions, GitLab CI, etc.
+     * @return the URI of the current build or null if it cannot be determined
+     */
+    public static String getBuildURI() {
+        return Optional.ofNullable(fromGithubActions()).orElseGet(() -> Optional.ofNullable(fromGitlabCI())
+                .orElseGet(() -> Optional.ofNullable(fromEnvironment(JENKINS_BUILD_URL))
+                        .orElseGet(() -> Optional.ofNullable(fromEnvironment(CIRCLE_BUILD_URL))
+                                .orElseGet(() -> Optional.ofNullable(fromEnvironment(TRAVIS_BUILD_WEB_URL))
+                                        .orElseGet(() -> Optional.ofNullable(fromEnvironment(DRONE_BUILD_LINK))
+                                                .orElse(null))))));
```
skhokhlov, line 62: For this and other methods in this class. Method must have `@Nullable` annotation if it can return `null`.

src/test/groovy/org/cyclonedx/gradle/PluginConfigurationSpec.groovy
```diff
@@ -690,4 +691,77 @@ class PluginConfigurationSpec extends Specification {
...
+                includeBuildSystem = true
+                buildSystemEnvironmentVariable = '${SERVER}/build/${BUILD_ID}'
+            }
+            dependencies {
+            }''', "rootProject.name = 'hello-world'")
+
+        when:
+        def result = GradleRunner.create()
+            .withProjectDir(testDir)
+            .withEnvironment(["SERVER" : "https://ci.example.com", "BUILD_ID" : "123"])
```
skhokhlov, line 753: Please add tests when variable from the string does not exist, e.g. `buildSystemEnvironmentVariable = '${WRONG_NAME}'`

README.md
```diff
@@ -65,6 +65,10 @@ cyclonedxBom {
     includeLicenseText = false
     // Include resolution of full metadata for components including licenses. Defaults to 'true'
     includeMetadataResolution = true
+    // Attempt to include the build-system URL by reading environment variables from common CI system such as GitHub Actions, GitLab CI, Drone, Jenkins, Travis CI, and Circle CI. Defaults to 'false'
+    includeBuildSystem = true
+    // if includeBuildSystem is true, the given environment variables will be used to construct the build-system URL that will be included in the BOM. Optional, defaults to `null`.
+    buildSystemEnvironmentVariable = '${CUSTOM_CI_URL}/jobs/${CUSTOM_JOB_ID}'
```
skhokhlov, line 71: Documentation must mention that braces `{}` are not optional here. Otherwise it's not clear, that `'$CUSTOM_CI_URL/jobs/$CUSTOM_JOB_ID'` will not work

skhokhlov (changes requested):
I like this feature. Left few considerations about the implementation.

src/test/groovy/org/cyclonedx/gradle/PluginConfigurationSpec.groovy
jeremylong: done

README.md
jeremylong: done

src/main/java/org/cyclonedx/gradle/utils/EnvironmentUtils.java
jeremylong: done

skhokhlov (approved):

PR #570, opened 2025-01-23
fix: manufacture. Reasoning: https://cyclonedx.org/docs/…

barblin:
…1.6/json/#metadata_manufacture deprecated. Use https://cyclonedx.org/docs/1.6/json/#metadata_manufacturer instead

### DIFF
--- a/src/test/groovy/org/cyclonedx/gradle/utils/OrganizationalEntityUtilTest.groovy
+++ b/src/test/groovy/org/cyclonedx/gradle/utils/OrganizationalEntityUtilTest.groovy
@@ -1,0 +1,137 @@
+package org.cyclonedx.gradle.utils
+
+import com.fasterxml.jackson.databind.ObjectMapper
+import org.cyclonedx.gradle.TestUtils
+import org.cyclonedx.model.Bom
+import org.gradle.testkit.runner.GradleRunner
+import org.gradle.testkit.runner.TaskOutcome
+import spock.lang.Specification
+
+class OrganizationalEntityUtilTest extends Specification {
+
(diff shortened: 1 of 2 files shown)


barblin:
Hey, we found that manufacture was empty if all fields are empty. We think the field should not be added in the SBOM at all if it contains no fields. 

Also, we found that you use https://cyclonedx.org/docs/1.6/json/#metadata_manufacture, which is deprecated. Probably https://cyclonedx.org/docs/1.6/json/#metadata_manufacturer is the appropreate replacement.

barblin:
@skhokhlov what do you think?

src/test/groovy/org/cyclonedx/gradle/utils/OrganizationalEntityUtilTest.groovy
```diff
@@ -0,0 +1,137 @@
...
+
+        then:
+        result.task(":cyclonedxBom").outcome == TaskOutcome.SUCCESS
+        File jsonBom = new File(testDir, "build/reports/bom.json")
+        Bom bom = new ObjectMapper().readValue(jsonBom, Bom.class)
+
+        assert bom.getMetadata().getManufacturer() == null
+    }
+
+    def "manufacturer should be empty if empty organizational entity is provided"() {
```
skhokhlov, line 95: Probably it `should not be empty`

barblin: @skhokhlov done :)

skhokhlov (approved):
@barblin thank you for the fix!

PR #580, opened 2025-02-03
fix: copy configurations into separate collection to avoid concurrent…

gordonrousselle:
When constructing a stream on the collection of existing configurations (i.e. `project.getConfigurations()`) it is possible to encounter a `ConcurrentModificationException`. This happens when another task updates this collection in parallel. To avoid this exception, the contents of the collection are copied before creating a stream.


skhokhlov:
Should we add some logging to print configuration and project name plugin is currently working on? It should provide insights for debugging

gordonrousselle:
> Should we add some logging to print configuration and project name plugin is currently working on? It should provide insights for debugging

Good idea. That's added.

skhokhlov (approved):

PR #656, opened 2025-07-31
build: Make org.cyclonedx:cyclonedx-core-java an api dependency

### LINKED ISSUE #655: Declare `org.cyclonedx:cyclonedx-core-java` as an API dependency
## Is your feature request related to a problem? Please describe.
The README describes how to modify SBOM metadata by using classes from `org.cyclonedx:cyclonedx-core-java`. Unfortunately this plugin declares and implementation dependency on `cyclonedx-core-java` which makes it inaccessible. If one wants to use those classes, they need to explicitly add `cyclonedx-core-java` dependency to Gradle build script.

## Describe the solution you'd like
Changing dependency declaration from `implementation` to `api` would make `cyclonedx-core-java` classes accessible in build script.

## Contribution 

- [x] I am willing to provide an implementation
- [ ] I will wait until somebody else implements it

zarebski-m:
Closes #655

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -23,7 +23,7 @@ repositories {
 }
 
 dependencies {
-    implementation("org.cyclonedx:cyclonedx-core-java:10.2.1") {
+    api("org.cyclonedx:cyclonedx-core-java:10.2.1") {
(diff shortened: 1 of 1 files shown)


skhokhlov (approved):
Thank you for the fix!

PR #670, opened 2025-08-20
Stop adding self references and dependency constraints as direct dependencies

### LINKED ISSUE #669: Dependency constraints incorrectly added as project direct dependency
## Describe the bug
Basically what was uncovered in https://github.com/spring-projects/spring-boot/issues/46924

## To Reproduce
New gradle project, add dependency constraints instead of  dependencies
```gradle
plugins {
	id 'java'
	id 'org.cyclonedx.bom' version '2.3.1'
}

group = 'com.example'
version = '0.0.1-SNAPSHOT'

repositories {
	mavenCentral()
}

dependencies {
	constraints {
		implementation 'org.jspecify:jspecify:1.0.0'
	}
	implementation 'com.google.guava:guava:33.4.8-jre'
}
```
run cyclonedxBom gradle task and you will find jspecify as a direct dependency of com.example
```json
{
  ...
  "dependencies" : [
    {
      "ref" : "pkg:maven/com.example/minimal-gradle-java-lib@0.0.1-SNAPSHOT?project_path=%3A",
      "dependsOn" : [
        "pkg:maven/org.jspecify/jspecify@1.0.0?type=jar",
        "pkg:maven/com.google.guava/guava@33.4.8-jre?type=jar"
      ]
    },
    ...
  ],
}
```
note that the constraint is not picked up as a direct dependency when they have no effect (eariler guava versions such as `com.google.guava:guava:32.0.1-jre` did not use jspecify), since they will not be included in the node tree.

## Expected behavior
Not including constraints as a dependency

## Environment  (please complete the following information):

- OS: Any (happens to be MacOS)
- Gradle version 9
- CycloneDX Plugin version 2.3.1

## Contribution 

tjog:
Fixes #669 

Would be great if this makes 3.0.0

### DIFF
--- a/src/main/java/org/cyclonedx/gradle/DependencyGraphTraverser.java
+++ b/src/main/java/org/cyclonedx/gradle/DependencyGraphTraverser.java
@@ -104,9 +104,12 @@ class DependencyGraphTraverser {
                     if (dep.isConstraint()) {
                         continue; // Skip constraints as they do not represent a dependency in the graph
                     }
-                    if (dep instanceof ResolvedDependencyResult) {   
+                    if (dep instanceof ResolvedDependencyResult) {
                         final ResolvedComponentResult dependencyComponent =
                                 ((ResolvedDependencyResult) dep).getSelected();
+                        if (graphNode.id.equals(dependencyComponent.getId())) {
+                            continue; // Skip self-references
+                        }
                         LOGGER.debug(
                                 "{} Node with ID {} has dependency with ID {}",
                                 LOG_PREFIX,
--- a/src/test/groovy/org/cyclonedx/gradle/PluginConfigurationSpec.groovy
+++ b/src/test/groovy/org/cyclonedx/gradle/PluginConfigurationSpec.groovy
@@ -1013,7 +1013,7 @@ class PluginConfigurationSpec extends Specification {
         where:
         taskName             | reportLocation
         "cyclonedxBom"       | "build/reports/cyclonedx"
-        // "cyclonedxDirectBom" | "build/reports/cyclonedx-direct" // TODO: enable once direct does not reference itself
+        "cyclonedxDirectBom" | "build/reports/cyclonedx-direct"
         javaVersion = JavaVersion.current()
     }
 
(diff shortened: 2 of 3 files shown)


src/main/java/org/cyclonedx/gradle/DependencyGraphTraverser.java
```diff
@@ -101,9 +101,15 @@ Map<SbomComponentId, SbomComponent> traverseGraph(
...
                 LOGGER.debug("{} Traversing node with ID {}", LOG_PREFIX, graphNode.id);
                 for (final DependencyResult dep : graphNode.getResult().getDependencies()) {
+                    if (dep.isConstraint()) {
+                        continue; // Skip constraints as they do not represent a dependency in the graph
+                    }
                     if (dep instanceof ResolvedDependencyResult) {
                         final ResolvedComponentResult dependencyComponent =
                                 ((ResolvedDependencyResult) dep).getSelected();
+                        if (graphNode.id.equals(dependencyComponent.getId())) {
+                            continue; // Skip self-references
```
tjog, line 111: Im not sure how these appear, but this was a difference discovered between the direct task and aggregate task.

Due to line 105 here https://github.com/CycloneDX/cyclonedx-gradle-plugin/blob/451288e33faa5b39e68c0c6de487764ba761fc2b/src/main/java/org/cyclonedx/gradle/CyclonedxAggregateTask.java#L100-L110

The self references are hidden from the aggregate bom, but why even include them in the first place?

This in turn forced changes to how the test below asserts how many dependencies the root has. I think the test case is a bit weird. But maybe there is a good explanation for adding self-references currently? (And testing it)

tjog:
I realize now after looking at open PRs that #641 incorporates the dependency constraint fix. Let me know if you want to prioritize that draft PR instead @skhokhlov

src/test/groovy/org/cyclonedx/gradle/PluginConfigurationSpec.groovy
```diff
@@ -902,4 +902,66 @@ class PluginConfigurationSpec extends Specification {
...
+        }
+        assert rootDependency != null
+        assert rootDependency.getDependencies().size() == 1
+        assert rootDependency.getDependencies().get(0).getRef().contains("guava@33.4.8-jre")
+        
+        // Verify jspecify constraint is still an entry in bom overall
+        def jspecifyDependency = bom.getDependencies().find { dep ->
+            dep.getRef().contains("jspecify@1.0.0")
+        }
+        assert jspecifyDependency != null
```
skhokhlov, line 952: I don't agree with this test case. Although `org.jspecify:jspecify:1.0.0` declared as a constraint here, it's still [a dependency of `com.google.guava:guava:33.4.8-jre`](https://github.com/google/guava/blob/v33.4.8/guava/pom.xml#L31-L34 ) and must appear in dependency graph.

skhokhlov: I agree. What I mean is can you please check here in the test that jspecify is still in the graph as a guava dependency?

tjog: Added the check in my most recent push

Issue #596, opened 2025-03-04
An error occurred attempting to extract POM from artifact


artn01:
## Describe the bug
I have a following gradle task in my build.gradle to generate SBOM:

```
plugins {
    id 'org.cyclonedx.bom' version '2.2.0'
}

cyclonedxBom {
    outputFormat = 'json'
    destination = file("sbom")
    outputName = 'myProject-bom'
    projectType = 'application'
}
```

So, when running the task, I keep getting the following error:
```
An error occurred attempting to extract POM from artifact
java.util.zip.ZipException: zip END header not found
```
However, the task executes successfully, and the SBOM is generated. I was able to deduct that this happens if SBOM was not previously generated in the `sbom/myProject-bom.json`. If there already is a file with the same name present on this location, error dissappears.

## To Reproduce
1. Copy the code mentioned in the issue in your `build.gradle` file.
 2. Run the task.

## Expected behavior
File is generated / rewritten (if file already exists on that location), no error is thrown and build is successful.

## Environment  (please complete the following information):

- OS: Windows 10
- Gradle version 8.6
- CycloneDX Plugin version 2.2.0

## Contribution 

- [ ] I am willing to provide a fix
- [x] I will wait until somebody else fixes it

artn01:
Seems that the following package causes the error, was able to find it while running in debug mode:

```
CycloneDX: Including meta data for node org.eclipse.jetty.orbit:javax.servlet:3.0.0.v201112011016
An error occurred attempting to extract POM from artifact
java.util.zip.ZipException: zip END header not found
```

wbiker:
I have the same issue on Windows 11, with gradle 8.10, 8.12 and kotlin.
In the debugger, I can see that the cyclonedx plugin is trying to open a pom file as a jar, causing this exception. It want to extract the metadata it seems and expecting to have jar files only here.

![Image](https://github.com/user-attachments/assets/80225b21-9844-4d3e-b0f2-790e37f2c9ea)

The issue happens in the MavenHelper class, in this method:
    @Nullable MavenProject extractPom(final File artifact, final ModuleVersionIdentifier mid) {
in line:
                final JarFile jarFile = new JarFile(artifact);
artifact is the pom file, see attached screenshot.

steven-sheehy:
I see this too but seems to be for:

[...]

PR #701, opened 2025-10-01
fix read pom files directly

lorenzbaier:
should fix https://github.com/CycloneDX/cyclonedx-gradle-plugin/issues/596


skhokhlov (changes requested):
Thank you for the pull request. We need to verify that the bug is actually fixed. Please add a test for that.

You can find [this test with local maven repository](https://github.com/CycloneDX/cyclonedx-gradle-plugin/blob/master/src/test/groovy/org/cyclonedx/gradle/DependencyResolutionSpec.groovy#L100-L144) helpful.

PR #758, opened 2026-01-21
fix: pass correct string representation when adding external references

### LINKED ISSUE #743: External references of components are always empty
### Bug Description

Starting with version `3.0.0`, all properties `$.components[*].externalReferences` seem to be empty, although present in the resulting `BOM` of each task. Both `cyclonedxBom` and `cyclonedxDirectBom` and XML and JSON output seem to be affected.

### Self-contained Reproducer Project

Run the test `DependencyResolutionSpec#only add component with valid purl` and inspect e.g. commons-codec, which specifies an issue tracker that is missing in the resulting XML and JSON.

### Expected Behavior

That `externalReferences` are filled with the values extracted from the POM.

### Gradle build scan URL (optional)

_No response_

### OS

Windows

### Gradle version

9.2.0

### CycloneDX Plugin version

[3.0.0, 3.1.0]

### Additional Context

Debugging shows that the external references are present in the `BOM`, but the type is always `null` which probably leads to them not being serialized. Digging a bit more shows that the type parsing is wrong:

https://github.com/CycloneDX/cyclonedx-gradle-plugin/blob/d28dcf64084862f524ce7ae48bebcf0d48d53509/src/main/java/org/cyclonedx/gradle/model/SbomMetaData.java#L77-L81

Here, the upper case enum value name is used as an argument to `org.cyclonedx.model.ExternalReference.Type#fromString`, but this method expects a lower case name, see e.g. here:

https://github.com/CycloneDX/cyclonedx-gradle-plugin/blob/d28dcf64084862f524ce7ae48bebcf0d48d53509/src/main/java/org/cyclonedx/gradle/model/SbomMetaData.java#L69

### Contribution

transcengopher:
* when adding external reference to output BOM, pass the same string representation (the type name) that will later be used to search the enumeration entry
* add test
  * add dummy scm information to one of the POMs in test local repo

closes #743

### DIFF
--- a/src/main/java/org/cyclonedx/gradle/model/SbomMetaData.java
+++ b/src/main/java/org/cyclonedx/gradle/model/SbomMetaData.java
@@ -66,7 +66,7 @@ public class SbomMetaData implements Serializable {
         metaData.setPublisher(component.getPublisher());
         if (component.getExternalReferences() != null) {
             component.getExternalReferences().forEach(reference -> {
-                metaData.addExternalReference(reference.getType().toString(), reference.getUrl());
+                metaData.addExternalReference(reference.getType().getTypeName(), reference.getUrl());
             });
         }
         return metaData;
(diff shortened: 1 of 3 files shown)


skhokhlov (approved):

skhokhlov:
Thank you for the fix!

PR #783, opened 2026-02-18
feat: exclude artifacts from sbom generation

### LINKED ISSUE #643: Exclude dependencies by name
Can we please have the possibility to exclude specific dependencies?

## Example:
```
...
dependencies {
    testImplementation platform('org.junit:junit-bom:5.9.1')
    implementation 'group.own.lib:own-lib:1.0.0'
}
...

cyclonedxBom {
    excludes = [
        'org.junit:junit-bom',
        'group.own.lib:own-lib'
    ]
}
/* or something like
cyclonedxBom {
    dependencies {
        exclude(dependency('org.junit:junit-bom'))
        exclude(dependency('group.own.lib:own-lib'))
    }
}
*/

...
```

bilak:
closes #643

Added naive implementation by which it could be possible to exclude artifacts from sbom generation.


skhokhlov (changes requested):
Thanks for the contribution, this is a useful feature and the overall structure is on the right track! There are a few issues worth addressing before merging — one bug, a design question about how the two tasks relate, and some implementation cleanups.

PR #785, opened 2026-02-23
feat: add support for configurable hash algorithms

ctran:
Implement the ability to only generate the hash/digest for specific algorithms instead of all available algorithms. On extremely large projects, the digest calculation can add a significant amount of time and space.

This is a follow up to @jeremylong change in https://github.com/CycloneDX/cyclonedx-core-java/pull/776.


README.md
```diff
@@ -135,23 +135,33 @@ import org.cyclonedx.model.Component
 
 tasks.named<CyclonedxDirectTask>("cyclonedxDirectBom") {
     projectType = Component.Type.APPLICATION
+>>>>>>> 271906c (feat: add support for configurable hash algorithms in SBOM generation)
```
jeremylong: @ctran this looks like a merge conflict marker was left in?
```suggestion
```

jeremylong (approved):
LGTM

src/main/java/org/cyclonedx/gradle/BaseCyclonedxTask.java
```diff
@@ -186,6 +186,18 @@ public abstract class BaseCyclonedxTask extends DefaultTask {
...
+     * The hash algorithms to include in the SBOM components.
+     * Supported values depend on the CycloneDX schema version.
+     * Common values: "MD5", "SHA-1", "SHA-256", "SHA-512", "SHA3-256", "SHA3-512".
+     * If not set, all algorithms supported by the schema version are included.
+     *
+     * @return the list of hash algorithms to include
+     */
+    @Input
+    @Optional
+    public abstract ListProperty<String> getHashAlgorithms();
```
skhokhlov: `hashAlgorithms` property is exposed on both tasks, but only direct one uses it. This property declaration should move to `CyclonedxDirectTask`

src/main/java/org/cyclonedx/gradle/SbomBuilder.java
```diff
@@ -79,11 +81,24 @@ class SbomBuilder<T extends BaseCyclonedxTask> {
...
+    private List<Hash.Algorithm> selectHashAlgorithms(final T task) {
+        final List<Hash.Algorithm> defaults = HashUtils.selectAlgorithms(this.version);
+        final List<String> configured = task.getHashAlgorithms().get();
+        if (configured.isEmpty()) {
+            return defaults;
+        }
+
+        final Set<Hash.Algorithm> requested =
+                configured.stream().map(Hash.Algorithm::fromSpec).collect(Collectors.toCollection(LinkedHashSet::new));
+        return defaults.stream().filter(requested::contains).collect(Collectors.toList());
```
skhokhlov, line 131: Configured algorithms can be silently dropped. It intersects user config with the plugin’s default list. It therefore accepts `Streebog-256` under schema 1.7, but passes an empty list to the Core and produces a successful SBOM with no component hashes. The config must either be honored or fail loudly; silent omission makes the SBOM contradict its configuration.

skhokhlov (changes requested):
Thank you for the update on this. I spotted a few small issues. Happy to merge once they fixed.

src/main/java/org/cyclonedx/gradle/SbomBuilder.java
ctran: Added validation and unit tests.

ctran:
> Thank you for the update on this. I spotted a few small issues. Happy to merge once they fixed.

@skhokhlov I've made the requested changes.  Thank you for your time and Have a great weekend!

PR #822, opened 2026-04-07
#821 - build fails with 'Cannot mutate when maven-publish attaches SBOM artifacts to publication

skhokhlov:


sergej-koscejev:
I have submitted #853 with a proposed fix for this problem.

PR #845, opened 2026-05-28
feat: make plugin compatible with isolated projects

JonasMorrissey:
I'm not sure how appropriate this pull request is - as much as I tried not to change anything, it does introduce one significant API break: the plugin would need to be applied on each (sub)project individually instead of just on the root. That alone might be a non-starter, and I'd completely understand if so. Please feel free to close :)

As to the content of this PR:
The current architecture is incompatible with both isolated projects because, from the root project, it iterates over subprojects, configures tasks on them, and then wires it all back together via configurations.

Instead, I think (hope) the "modern"/suggested gradle way is to use a build service. At configuration time, every project registers the following information inti into this service:
* the output(s) of its directBomTask 
* the task path (e.g. :app-a:cyclonedxDirectBom), used for dependsOn wiring, 
* and the project path (e.g. :app-a), used only so that we can have aggregate bom tasks on non-root projects. (Otherwise the test "should run aggregate task on root and subproject" would fail)

Then the aggregate task can simply pull its inputs from this service without cross-referencing any other projects.

### DIFF
--- a/.github/workflows/ci-build.yaml
+++ b/.github/workflows/ci-build.yaml
@@ -24,7 +24,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout repository
-        uses: actions/checkout@v6.0.2
+        uses: actions/checkout@v6.0.3
 
       - name: Set up JDK
         uses: actions/setup-java@v5.2.0
--- a/.github/workflows/ci-publish.yaml
+++ b/.github/workflows/ci-publish.yaml
@@ -20,7 +20,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout repository
-        uses: actions/checkout@v6.0.2
+        uses: actions/checkout@v6.0.3
         with:
           ref: ${{ github.event.release.tag_name }}
 
--- a/.github/workflows/dependency-submission.yaml
+++ b/.github/workflows/dependency-submission.yaml
@@ -14,7 +14,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout repository
-        uses: actions/checkout@v6.0.2
+        uses: actions/checkout@v6.0.3
       - name: Setup Java
         uses: actions/setup-java@v5.2.0
         with:
--- a/.github/workflows/gradlew-update.yaml
+++ b/.github/workflows/gradlew-update.yaml
@@ -15,7 +15,7 @@ jobs:
     if: ${{ github.repository == 'CycloneDX/cyclonedx-gradle-plugin' }}
     steps:
       - name: Checkout repository
-        uses: actions/checkout@v6.0.2
+        uses: actions/checkout@v6.0.3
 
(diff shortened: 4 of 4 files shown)


skhokhlov (changes requested):
Thanks for this - and for being upfront about the API-break tradeoff in the description. Your diagnosis is exactly right: the root-iterates-subprojects pattern is fundamentally incompatible with Project Isolation, and per-project application is the correct direction. I genuinely appreciate the effort here; this pushed a change I'd been putting off.

I don't want to merge this implementation, though, and I want to be transparent about why so the work isn't wasted.

**The direction is right; the mechanism isn't.** Routing aggregation through a shared mutable build service has two problems that undercut the very goal of the PR:

1. **Configuration cache.** The service accumulates its registrations at configuration time, but that state isn't part of the service's parameters, so it isn't restored on a configuration-cache hit. The aggregation inputs are populated as a side effect of configuration that won't re-run.
2. **Isolation ordering.** The aggregate task reads the registry when the task graph is built, which assumes every subproject has already been configured. Project Isolation and configuration-on-demand specifically *remove* that guarantee - so under the feature this PR targets, the registry can be incomplete. The mechanism contradicts its own goal.

The plugin already uses variant/artifact-based aggregation (an intentional 2.x -> 3.x design), and that's the right foundation - it's what Gradle's own `jacoco-report-aggregation` / `test-report-aggregation` plugins do, and it's what the configuration cache is built to serialize. The only thing that actually breaks Isolation today is that the wiring lives on the root and reaches into subprojects. The fix is to move that wiring into each project's own application path and keep the variant-based aggregation - no build service needed.

I've split the target design into two issues:

- #847 - Project Isolation support (the goal + acceptance bar)
- #848 - migration to an explicit, opt-in aggregator setup (the mechanism)

Two things from your PR I want to explicitly keep: the per-project application model, and having the **aggregate task own the root->member dependency edges** (your added edge synthesis is the right idea - under explicit membership it becomes intentional rather than inherited from cross-project traversal).

A couple of things that shouldn't carry over: the tolerant BOM parsing (a missing/unparseable member should make the aggregate **fail loudly**, not silently under-reporting an SBOM is the worst outcome for a security-consumed artifact), and anything that lets the aggregate silently include or exclud

If you're up for reworking it along the lines in those issues - per-projec explicit `cyclonedxAggregation project(...)` membership, a version-gated Isolation functional test, and loud failure - I'd be happy to guide it through. No pressure either way, and thank you again for surfacing this.

JonasMorrissey:
Thanks a lot for looking at my PR and for your feedback! 😊 You are absolutely right about the drawbacks of using a build service. In our setup, we always call gradlew cyclonedxBom on the root project, which then recursively runs the task in each subproject, so we weren't affected by the second issue. But of course if somebody were to run gradlew **:**cyclonedxBom, in some constellations, they could potentially get a much smaller bom - silently! 

I'll be happy to have a go at #848 going in the direction you outlined.

PR #846, opened 2026-06-02
Fix "Cannot mutate" Gradle error when trying to add SBOM output to artifacts

czarnecki:
Fixes https://github.com/CycloneDX/cyclonedx-gradle-plugin/issues/821.

The problem seems to have been that `configure` was called after the project was evaluated, which caused issues when adding custom artifacts.

To fix this I use `flatMap` and `isPresent` to check if tasks in the sub projects are enabled or if `jsonOutput` or `xmlOutput` is set.
I didn't use the `filter` method available on `Provider` because that is only available in Gradle 8.5+ and still an incubating feature. Additionally I had to use `flatMap` with nested `Provider` because returning `null` directly in the `Transformer` caused a false positive in the NullAway (https://github.com/uber/NullAway/issues/1532) to trigger.
The workaround can me simplified once the bug has been fixed, or removed once Gradle 8.5+ is the minimum supported version

I added a reproducer for the setup in my project that caused the error to happen. I've also tested the fix in my project and it no longer occurs there.

### DIFF
--- a/src/main/java/org/cyclonedx/gradle/CyclonedxPlugin.java
+++ b/src/main/java/org/cyclonedx/gradle/CyclonedxPlugin.java
@@ -21,6 +21,7 @@ package org.cyclonedx.gradle;
 import com.google.common.collect.ImmutableMap;
 import java.util.stream.Stream;
 import javax.inject.Inject;
+import org.gradle.api.DefaultTask;
 import org.gradle.api.JavaVersion;
 import org.gradle.api.Plugin;
 import org.gradle.api.Project;
@@ -30,6 +31,7 @@ import org.gradle.api.file.Directory;
 import org.gradle.api.logging.Logger;
 import org.gradle.api.logging.Logging;
 import org.gradle.api.provider.Provider;
+import org.gradle.api.specs.Spec;
 import org.gradle.api.tasks.TaskProvider;
 
 /**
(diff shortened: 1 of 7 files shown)


src/main/java/org/cyclonedx/gradle/CyclonedxPlugin.java
```diff
@@ -104,51 +106,62 @@ public void apply(final Project project) {
...
-                                                        evaluatedProject.getPath(),
-                                                        "configuration",
-                                                        cyclonedxDirectConfigurationName)));
-                    });
+                                        cyclonedxDirectConfigurationName,
+                                        directTask.map(BaseCyclonedxTask::getXmlOutput),
+                                        a -> a.builtBy(directTask));
+                    }
+
+                    if (filterProvider(project, directTask, task -> task.getJsonOutput()
```
skhokhlov: The `filterProvider(...).isPresent()` pattern reads as lazy because it's `Provider`-shaped, but `.isPresent()` forces the source `TaskProvider` to resolve, so the `cyclonedxDirectBom` task is realized during `afterEvaluate` regardless. I verified this with a marker in the task's registration action, `./gradlew help` realizes the direct task for every project (root and both subprojects) on this branch, but 0 times on master's `directTask.configure(...)`. So the helper doesn't avoid realization - it causes it.

Since the task is realized anyway, the provider, the `Spec/DefaultTask` imports, and the `project.provider(() -> null)` NullAway workaround can all go. Realizing once with `.get()` is behaviorally identical and far clearer - and lets the artifact value be the clean single-level `RegularFileProperty` instead of a nested `Provider<Provider<RegularFile>>`:

```java
 final CyclonedxDirectTask directTask = evaluatedProject
         .getTasks()
         .named(cyclonedxDirectTaskName, CyclonedxDirectTask.class)
         .get();

 if (!directTask.getEnabled()) { /* log skip; return; */ }

 if (directTask.getXmlOutput().isPresent()) {
     evaluatedProject.getArtifacts().add(
         cyclonedxDirectConfigurationName, directTask.getXmlOutput(), a -> a.builtBy(directTask));
 }
 // …same for getJsonOutput()…
```


Could you also add a short comment on why the wiring must be eager here (the outgoing config can be consumed before the task realizes #821), so this doesn't get "optimized" back into a deferred configure block later?

src/test/groovy/org/cyclonedx/gradle/DependencyResolutionSpec.groovy
```diff
@@ -594,6 +594,25 @@ class DependencyResolutionSpec extends Specification {
...
+
+        when:
+        def result = GradleRunner.create()
+            .withProjectDir(testDir)
+            .withArguments(TestUtils.arguments("publish"))
+            .withPluginClasspath()
+            .build()
+
+        then:
+        result.task(":publish").outcome == TaskOutcome.SUCCESS
```
skhokhlov, line 613: The test asserts only `esult.task(":publish").outcome == SUCCESS`. Because the fix changes the artifact to a provider, a future regression that silently drops the SBOM (empty/absent provider) would still publish successfully and keep this test green. Let's lock that in by asserting the published artifact actually contains both SBOM entries and assert entries present and non-empty.

skhokhlov (changes requested):
Thanks for the PR! Just a few suggestions here how to make it cleaner and I'll be happy to merge.

Also just a head up about PluginConfigurationSpec:`should not realize cyclonedx tasks for unrelated builds`. It asserts `result.task(":cyclonedxDirectBom") == null` which stays green even though this PR now realizes the task at config time. Since that's an accepted trade-off here (realization is construction-only; no dependency resolution), but the test name overstates what it checks. I suggest either renaming it to reflect that it guards execution/resolution (not construction), or strengthening it to assert the property that genuinely still matters now that we realize eagerly: on `help`, no dependency resolution runs and no SBOM output file is produced.

czarnecki:
Thanks for the review!
I've expanded the test I added to verify that the Zip and the files exist.
I've also renamed the test you've mentioned.
In regards to the realization of the task:
I've come across this issue https://github.com/gradle/gradle/issues/35397 which mentions that `ArtifactHandler#add` is eager under the hood. In this issue they also mention a different way with which you should add artifacts, which I've used now.
With this `cyclonedxDirectBom` seems to no longer be configured when `help` is run, unless I've made a mistake in my testing.
I've also removed the `filterProvider` method I had added and simply just call `get()` to get the task. This should be fine since this is now within lazy configuration, which should only be called when the configuration is wanted, in which case we also want the tasks anyway

I hope these changes are fine, I'm also happy to make further adjustments

sergej-koscejev:
I propose a slightly different approach in #853, using `artifacts(Provider<Iterable>)`. This makes it possible to drop afterEvaluate completely and achieve correctness in more cases because of more laziness.

skhokhlov:
I agree that approach in #853 is suits this plugin more. The `afterEvaluate` is rather a hack than a correct approach, so removing it should remove source of the issue instead making fixes around it.

czarnecki:
Yea I agree, I'll close this PR then

PR #853, opened 2026-06-10
fix: publish direct SBOMs lazily

sergej-koscejev:
This should fix issue #821.

The solution is to replace afterEvaluate with lazy (provider) API. This opens a few minor cans of worms that I could, I believe, successfully close with the help of Claude.

## The fix (commit 1)

- Register the BOM output files as outgoing artifacts at plugin apply time via the task provider (`taskProvider.map(...)`), so they are visible without realizing or running the task.
- Add the aggregate configuration's project dependencies through `DependencySet.addAllLater` with a provider that filters out projects whose direct BOM task is missing or disabled. This preserves the enabled-check of the old `afterEvaluate` block (a disabled subproject stays out of the aggregate and out of the root project's BOM) while keeping the wiring lazy, so unrelated builds still don't realize any CycloneDX tasks.

## Ignoring missing SBOMs

Commit 1 alone will simply skip SBOMs that were not generated, assuming it was due to a disabled task. If the SBOM turns out to not be generated due to an error, we would hide this error (there would be a warning but who reads warnings...). Therefore:

## Proposal: fail on missing input SBOMs (commit 2, separable)

To fix the problem we make `CyclonedxAggregateTask` fail with an explanatory error when an input SBOM is declared but its file was never generated. This can only happen when the producing task was skipped at execution time (e.g. via `onlyIf`), since disabled tasks are already excluded from the configuration. The omission must be made explicit by disabling the producing task. (Previously such files failed the task with an unhelpful `ParseException`.)

This commit is independent of the fix and can be dropped if the behavior change is not wanted.

## Flaky test fix (commit 3)

Diagnosed and fixed by Claude:

> Disable file system watching in functional tests
>
> Tests that rewrite a build script and immediately run the next build in a reused test-kit daemon could see the configuration cache falsely reused, because the daemon's file-system-watching VFS had not yet processed the change. The task under test was then UP-TO-DATE instead of re-running, failing the test. Disabling watching makes the daemon re-hash build scripts from disk at each build start.

### DIFF
--- a/src/main/java/org/cyclonedx/gradle/CyclonedxAggregateTask.java
+++ b/src/main/java/org/cyclonedx/gradle/CyclonedxAggregateTask.java
@@ -33,6 +33,7 @@ import org.cyclonedx.model.BomReference;
 import org.cyclonedx.model.Component;
 import org.cyclonedx.model.Dependency;
 import org.cyclonedx.parsers.BomParserFactory;
+import org.gradle.api.GradleException;
 import org.gradle.api.file.ConfigurableFileCollection;
 import org.gradle.api.logging.Logger;
 import org.gradle.api.logging.Logging;
@@ -73,6 +74,9 @@ public abstract class CyclonedxAggregateTask extends BaseCyclonedxTask {
         final Bom aggregateBom = getRootProjectBom();
         final Map<String, Component> componentsByBomRef = new TreeMap<>();
         final Map<String, Set<String>> dependenciesByBomRef = new TreeMap<>();
+
+        checkForMissingInputSboms(files);
+
         for (final File subProjectBomFile : files) {
             final Bom subProjectBom =
                     BomParserFactory.createParser(subProjectBomFile).parse(subProjectBomFile);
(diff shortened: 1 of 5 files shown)


src/main/java/org/cyclonedx/gradle/CyclonedxAggregateTask.java
```diff
@@ -48,9 +50,25 @@ public abstract class CyclonedxAggregateTask extends BaseCyclonedxTask {
...
+     * Whether to ignore input SBOMs whose file does not exist instead of failing the task.
+     * An input SBOM is missing when its producing task was skipped at execution time (e.g. via onlyIf)
+     * so the SBOM was never generated. Note that projects whose direct BOM task is disabled are already
+     * excluded from the aggregation at configuration time and are not affected by this property.
+     * If not set, it defaults to false.
+     *
+     * @return true if missing input SBOMs should be ignored, false if they should fail the task
+     */
+    @Input
+    public abstract Property<Boolean> getIgnoreMissingInputSboms();
```
skhokhlov: Providing a clear, actionable `GradleException` is a great improvement for missing SBOMs. However, I believe it's better to drop the `ignoreMissingInputSboms` opt-out and always fail. A missing input SBOM means a subproject silently vanishes from the aggregate in this case. The BOM is a public-contract artifact, and silent under-reporting is worse than a loud break. The opt-out introduces the "warning nobody reads" path. Intentional exclusion already has a clean option: disable the producer's `cyclonedxDirectBom` task (`enabled = false`), which drops it from aggregation at configuration time.

skhokhlov (commented):
Thanks @sergej-koscejev  this is a solid, well-scoped fix. I have a suggestion for `ignoreMissingInputSboms` property, I'm happy to discuss it.

src/main/java/org/cyclonedx/gradle/CyclonedxAggregateTask.java
sergej-koscejev: I don't think it's a silent failure because it requires the build author to set the property to a non-default value, so I'd assume they know what they are doing in this case. But I'm leaving the property out, you can add it later if users ask for an opt-out.

sergej-koscejev:
@skhokhlov I think this issue is waiting for your decision now, from my point of view it can be merged. Let me know if I'm missing something.

skhokhlov (approved):

skhokhlov:
Thank you @sergej-koscejev for the PR and sorry for the delay with review!

PR #882, opened 2026-07-22
feat: configurable testConfigs for cdx:maven:package:test

luczus:
## Summary
- Add `testConfigs` on `cyclonedxDirectBom` so configuration names can be classified as Test Configurations when setting `cdx:maven:package:test`.
- Default remains `['^test.*']` (equivalent to the previous `startsWith("test")` check), so the SBOM Output Contract is unchanged without an opt-in (ADR 0004 / ADR 0005).
- Projects with custom test source sets (e.g. `e2eTest`) can override with patterns such as `['(?i).*test.*']`.

## Context
Today a component is labeled `cdx:maven:package:test = true` only when every in-scope configuration name starts with the literal `"test"`. That misses custom source sets such as `e2eTest` / `integrationTest`, so test-only tooling can appear as production dependencies in the Direct SBOM.

`includeConfigs` / `skipConfigs` already control which configurations are *scanned*; this change only affects the *label*. Classification stays name-pattern based (not derived from Gradle source sets) — see ADR 0005.

## Test plan
- [x] `./gradlew spotlessApply testJava21 --tests "*TestConfigurationLabelingSpec"`
- [ ] CI matrix on the PR
- [ ] Manual check: default build still marks `testImplementation` deps as test
- [ ] Manual check: `e2eTest` + `testConfigs = ['(?i).*test.*']` marks those deps as test

Made with [Cursor](https://cursor.com)

### DIFF
--- a/CONTEXT.md
+++ b/CONTEXT.md
@@ -31,6 +31,12 @@ The public compatibility surface of produced **Direct SBOMs**, **Aggregate SBOMs
 versioned with the plugin according to [ADR 0004](docs/adr/0004-version-the-sbom-output-contract-with-the-plugin.md).
 _Avoid_: Internal output, report format
 
+**Test Configuration**:
+A Gradle configuration treated as test evidence when labeling a component in a **Direct SBOM**. A component is marked
+test only when every configuration that contributed it is a **Test Configuration**. Classification is by configurable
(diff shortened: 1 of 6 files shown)


skhokhlov (approved):
Thank you for this!

Issue #883, opened 2026-07-24
Provide a stable task DSL independent of CycloneDX Core in 4.0


skhokhlov:
## User problem

Several task properties currently expose `cyclonedx-core-java` model and enum types directly. This makes ordinary
plugin configuration depend on the implementation library's API and turns core upgrades into migration work for
plugin users.

Users need a configuration surface that is natural in both Groovy and Kotlin, remains stable when CycloneDX Core
evolves, and offers a gradual migration from existing 3.x builds.

## Desired outcome

CycloneDX Gradle Plugin 4.0 provides a plugin-owned, Gradle-native DSL directly on the Direct and Aggregate SBOM tasks.
Users can configure the same effective metadata available through the existing core-backed properties without
constructing CycloneDX Core objects.

The 4.0 experience should have these properties:

- The canonical DSL uses plugin-owned terms and types, with validated string aliases where they improve Groovy and
  Kotlin usability.
- It has capability parity with every effective field reachable through the five existing core-backed properties,
  rather than attempting to mirror the complete CycloneDX object model.
- Existing 3.x configurations using `schemaVersion`, `projectType`, `organizationalEntity`, `licenseChoice`, and
  `externalReferences` remain functional throughout 4.x as a deprecated compatibility facade. Removal is no earlier
  than 5.0.
- Canonical and legacy configuration describe one result: equivalent duplicate values are accepted and conflicts fail
  with actionable diagnostics rather than silent precedence.
- Migration guidance is mechanical and example-driven for both Groovy and Kotlin. Detectable legacy use produces at
  most one migration warning per build.
- The canonical DSL does not expose a raw CycloneDX Core escape hatch.
- CycloneDX 1.7 becomes the default SBOM schema in 4.0 after support is delivered by #832. Explicitly selecting 1.6
  remains supported for users who need the previous output contract.

Apart from explicitly documented 4.0 output-contract changes, equivalent legacy and canonical configuration must
produce equivalent parsed CycloneDX meaning. Serializer formatting and irrelevant collection ordering are not
compatibility guarantees.

## Proposed DSL shape

The following examples preserve the direction discussed during planning. They are a reference for the task-local
hierarchy and migration experience, not a commitment to every identifier before Groovy and Kotlin compilation
fixtures validate the final API.

Kotlin, using plugin-owned typed values:

```kotlin
tasks.cyclonedxDirectBom {
    bom {
        schemaVersion(SchemaVersion.VERSION_17)
[...]

PR #890, opened 2026-07-28
feat: support user-declared properties on metadata and main BOM component objects

### LINKED ISSUE #706: Add support for properties
### Is your feature request related to a problem? Please describe.

I do have issues, but they're very specific to me. More broadly my problem is just _how can I add arbitrary data to further describe my component in the SBOMs_?

Things like somewhere to put a URL to a build server (although the standard has a specific spot for this), a commit hash, an EOS/EOL date, the dependency scopes included in this SBOM, or any other wild and wonderful things people might find they want/need in their SBOMs.

### Describe the solution you'd like

As of 1.3, the CycloneDX format supports a `properties` array in any `component`, including the metadata's component. I'd love the ability to just specify a map as a task input which gets used as the metadata component's properties.

That should let anybody do something simple like this:
```
tasks.cyclonedxBom {
  properties = ["foo": "bar"]
}
```

or more complicated stuff that takes outputs from other tasks and uses them instead in addition to a static collection.

### Describe alternatives you've considered

_No response_

### Additional context

There's a similar issue open with the Maven plugin, https://github.com/CycloneDX/cyclonedx-maven-plugin/issues/131.

Implementing this for Gradle seems much easier however.

### Contribution

- [ ] I am willing to provide an implementation
- [x] I will wait until somebody else implements it

transcengopher:
Implements support for defining properties in following BOM paths:

- `BOM.Metadata.Properties` via repeatable `metadataProperty {}` configuration block
- `BOM.Metadata.Component.Properties` via repeatable `componentProperty {}` configuration block

In every property block, `name` is a required property (eagerly resolved), while value is optional (lazily resolved)

API structure is somewhat inspired by discussion in #883, possibly would require minimal migration effort after proper DSL introduction in next major version.

closes #706


MalickBurger:
Hi @transcengopher thank you for your contribution. We will provide feedback as soon as we can

PR #896, opened 2026-08-05
docs: match includeConfigs to corresponding comment

chkpnt:
`compileClasspath` is not a runtime configuration.


MalickBurger:
Hi @chkpnt thank you for your contribution. We will provide feedback as soon as we can

PR #904, opened 2026-08-07
fix: prefer a declared license url over its name when the two disagree

JonasVautherin:
I had the issue where the name was "BSD" and it resolved to BSD-4-Clause where the actually licence was BSD-2-Clause. The URL was correct.

It feels to me that the URL should be preferred, because it is less ambiguous than the name.


MalickBurger:
Hi @JonasVautherin thank you for your contribution. We will provide feedback as soon as we can.

PR #922, opened 2026-09-01
fix: report what the SBOM could not describe

### LINKED ISSUE #921: A component whose metadata could not be read is reported as one that declares none
### Bug Description

When the plugin cannot read a component's POM, the component still appears in the Direct SBOM, with no licenses, no
publisher, and nothing in the document to say that anything went wrong. That is byte-for-byte what a component which resolves and
declares nothing looks like, so a consumer of the document cannot tell the two apart.

Three of the four paths that produce this are silent: `MavenProjectLookup.buildMavenProject` returns `null` when the
artifact resolution query yields no component, when the component carries no POM artifact, and when the artifact result
is an `UnresolvedArtifactResult`. The last of these discards a `getFailure()` that says exactly what went wrong. The
fourth path, where the POM or the effective model fails to parse, logs at error level and then returns `null` as well.

Logging is not the answer to this on its own. The build succeeds, so nothing prompts anyone to read its log, and where
build output is retained at all it is commonly dropped for successful runs, so by the time a consumer notices the
missing license the log is gone. The SBOM is the artifact that travels, and it carries no hint that anything failed.

Our setup is a Jenkins fleet with many workers, each with its own Gradle dependency cache, resolving through a
Sonatype Nexus proxy in front of Maven Central, and sharing a Develocity remote build cache. One worker could not read
the metadata of a component and produced a Direct SBOM in which an ordinary, properly licensed library carried no
licenses. Because `cyclonedxDirectBom` is cacheable, that document went to the shared cache, and every
later build whose inputs matched was served it, on workers and laptops whose own dependency caches were entirely healthy. A license
policy task consuming the document then failed like this:

```
> Task :<project>:updateThirdPartyLicenses FAILED
  No license found for 'ch.qos.logback:logback-classic:1.6.0'
```

That component declares no licenses of its own; they live in `ch.qos.logback:logback-parent`. So the message names a
component that is entirely fine, on a machine that is entirely fine, for a lookup that failed somewhere else hours
earlier and left nothing in the document to say so. It took two days and around sixty failed builds to find the machine
that produced it.

### Self-contained Reproducer Project

https://github.com/Donnerbart/cyclonedx-metadata-reproducer

```bash
./gradlew cyclonedxDirectBom
cat build/reports/cyclonedx-direct/bom.json
```

It resolves nine components from a file repository committed alongside it, so it needs no network beyond the plugin
itself. One component is healthy and the other eight are each broken differently. The build succeeds, and the document
reports:

| component             | what is wrong with it                        | licenses in the SBOM | logged |
|-----------------------|----------------------------------------------|----------------------|--------|
| `child-ok`            | nothing, its license comes from its parent   | `Apache-2.0`         | -      |
| `jaronly`             | the repository holds no POM for it           | none                 | no     |
| `badpom`              | its POM is an HTML error page                | none                 | yes    |
| `emptypom`            | its POM is zero bytes                        | none                 | yes    |
| `orphanpom`           | its parent is not in the repository          | none                 | yes    |
| `badparentchild`      | its parent's POM is an HTML error page       | none                 | yes    |
| `importsbadbom`       | the BOM it imports is an HTML error page     | none                 | yes    |
| `importsmissingbom`   | the BOM it imports is not in the repository  | none                 | yes    |
| `ownlicensebadparent` | declares MIT itself, its parent is broken    | none                 | yes    |

Eight causes, one indistinguishable result. `jaronly` is the case our fleet hit, and it writes nothing to the log at
all. `ownlicensebadparent` is worth a second look: it declares MIT in its own POM, and since Maven replaces the
inherited license list instead of merging into it, that declaration is complete without the parent. The effective model
fails anyway and the license the plugin had already read is dropped.

Every one of the eight looks like this in the document, with hashes elided:

```json
{
  "type": "library",
  "bom-ref": "pkg:maven/com.test/jaronly@1.0.0?type=jar",
  "group": "com.test",
  "name": "jaronly",
  "version": "1.0.0",
  "purl": "pkg:maven/com.test/jaronly@1.0.0?type=jar",
  "modified": false,
  "properties": [ { "name": "cdx:maven:package:test", "value": "false" } ]
}
```

Running the artifact resolution query for `jaronly` by hand shows what the plugin had in front of it and discarded:

```
com.test:jaronly:1.0.0 -> DefaultUnresolvedArtifactResult:
    Could not find jaronly-1.0.0.pom (com.test:jaronly:1.0.0).
```

### Expected Behavior

A component whose metadata could not be read should be distinguishable, in the document, from one that resolved and
declares no license. The build does not have to fail and no license should be invented. The document only has to carry
the fact that the lookup failed, so that whoever consumes the SBOM can act on it.

Two narrower expectations follow from the reproducer:

- A license the plugin has already read should survive. `ownlicensebadparent` declares MIT in its own POM, and since
  Maven replaces the inherited license list instead of merging into it, that declaration is complete without the parent.
  It is dropped today.
- The failure Gradle reported should not be discarded. For `jaronly` the artifact resolution query returns an
  `UnresolvedArtifactResult` whose `getFailure()` says exactly what went wrong, and the plugin returns `null` without
  logging or recording it.

### Gradle build scan URL (optional)

_No response_

### OS

Linux

### Gradle version

9.7.1

### CycloneDX Plugin version

3.4.1

### Why nothing below the plugin catches this

We have since traced our own occurrence through the proxy logs, and no layer below the plugin was in a position to
report it.

Our build fleet leaves through two egress addresses, and Maven Central had been rate limiting them. We put the
caching proxy back in front of the fleet to consolidate that traffic; it came up on the same egress address and was
throttled from its first request onward, for about an hour. The proxy presented each `429` to the client as `404`, and
logged nothing. In one second the worker that later produced the document was told that four POMs, including the one
behind the failure, do not exist:

```
client   HEAD /repository/maven-central/ch/qos/logback/logback-classic/1.6.0/logback-classic-1.6.0.pom        404
upstream HEAD https://repo1.maven.org/maven2/ch/qos/logback/logback-classic/1.6.0/logback-classic-1.6.0.pom   429
```

For Gradle a `404` is a legitimate answer that the repository carries no such artifact, and the metadata cache stores
that absence like any other result. The worker did not ask again for 24 hours, and every lookup in between was answered
from the record with no request and no error. That is the state this plugin then meets: an
`ArtifactResolutionQuery` that yields nothing, instantly, for a component whose jar resolved and whose POM is fine
upstream. The dependency cache holds the answer a repository gave, not a fact about the artifact, and once a negative
answer is in there it is indistinguishable from an artifact that was never published.

The proxy in question is a stock Sonatype Nexus on default settings, the layer whose entire purpose is to keep builds
insulated from upstream trouble, and it is the layer that turned an hour of throttling into a day of "this artifact does
not exist". We reproduced it afterwards on a clean instance of the same version with nothing configured beyond the proxy
repository, so this is how the product behaves, not something we misconfigured.

This will get more common. Maven Central's [FAQ on 429 responses](https://central.sonatype.org/faq/429-error/) says
"those thresholds will move downward", and the remedy it recommends is "caching artifacts, consolidating requests
through a repository manager", which puts more builds behind the kind of proxy that produced this. The
[joint statement](https://openssf.org/blog/2025/09/23/open-infrastructure-is-not-free-a-joint-statement-on-sustainable-stewardship/)
it links, signed by the stewards of Maven Central, PyPI, crates.io and RubyGems among others, extends that pressure past
Maven. A build cannot audit what its proxy chose to answer, so the document is where the distinction has to survive.

Content is not verified either. From the dependency caching guide, under
[Artifact reuse](https://docs.gradle.org/current/userguide/dependency_caching.html#sec:cache-artifact-reuse):

> Before downloading an artifact, Gradle attempts to retrieve the artifact's checksum by downloading an associated
> `.sha512`, `.sha256`, `.sha1`, or `.md5` file (attempting each in order). If the checksum is available, Gradle skips
> the download if an artifact with the same ID and checksum already exists. However, if the checksum cannot be retrieved
> from the remote server, Gradle proceeds to download the artifact but will ignore it if it matches an existing one.

The checksum decides whether a download can be skipped, and no documented step compares what arrived against it, which
is how a proxy's error page ends up stored as a POM.
[gradle/gradle#26542](https://github.com/gradle/gradle/issues/26542) is that case, a self-hosted proxy answering with an
HTML page or zero bytes, closed by improving the error message and leaving the behavior as it was.
[gradle/gradle#34158](https://github.com/gradle/gradle/issues/34158) asks that an artifact be discarded when its
checksum cannot be retrieved, and sits in the relevant team's backlog.

We have seen all of these shapes across the fleet: a POM the repository denies having, a POM stored as zero bytes, and a
POM whose first line is an HTML error page. The reproducer carries one component for each, alongside the unresolvable
and unreadable parent cases, because they take different paths inside this plugin and arrive at the same component
entry.

### Proposed fix

A component whose metadata could not be read carries a `cdx:gradle:metadata:unresolved` property naming what happened,
alongside the `cdx:maven:package:test` property the plugin already emits, and the reason Gradle or Maven reported goes
to the log at warning level:

| value                | what happened                                                    |
|----------------------|------------------------------------------------------------------|
| `pom-unresolved`     | the repositories returned no usable POM artifact                 |
| `pom-unparseable`    | a POM was returned and could not be read                         |
| `parent-unresolved`  | the effective model failed, and the component declares a parent  |
| `parent-unparseable` | as above, where what failed was obtained but unusable            |
| `model-incomplete`   | the effective model failed, and the component declares no parent |

No license is invented, the task does not fail by default, and healthy builds produce unchanged documents. Two details
shape the fix:

- **What a POM declares is kept even when the effective model fails.** Maven replaces the inherited list instead of merging into it, so a
  component that states its own licenses inherits none, and those statements are complete without the model. Discarding them
  because a parent or an import could not be read, which is what happens today, loses information the build had in hand.
- **A failing import cannot be told from a failing parent** for a component that declares a parent, because Maven does
  not report which input failed. Resolving the parent a second time on the failure path would settle it; the proposal
  does not do that, and records `parent-*` in that case, which over-attributes to the parent.

Emitting `NOASSERTION` was considered and rejected: the string appears in neither `bom-1.6.schema.json` nor
`bom-1.7.schema.json` as shipped by `cyclonedx-core-java` 13.1.0, and it would state a license the component does not
claim. Failing the build by default was rejected for the reason given below: a component that carries no POM is
legitimate in an artifact-only or Ivy repository, and reaches this code on the same path.

Under ADR 0004 this is optional metadata valid under the configured schema, added while existing fields, component
identities and dependency relationships keep their meaning, so it fits a minor release.

### A second way the document understates the build

A declared dependency that does not resolve never reaches the document at all. `DependencyUtils` configures the artifact
view with `lenient(true)`, so the module is dropped, `DependencyGraphTraverser` logs it at `info`, and the task
succeeds. A build declaring one dependency that cannot be resolved produces an SBOM with zero components and no
indication that anything is missing.

The proposal raises that line to `warn`, since the document cannot carry the fact. Putting a placeholder component in
the SBOM instead would change the default set of included components, which ADR 0004 reserves for a major version, and
would sit against ADR 0001, where the resolved graph is the authority for what the document contains. That question is
worth its own issue.

### Choosing to fail

Neither case fails the build by default, because a component with no POM at all is legitimate in an artifact-only or
Ivy repository and arrives here exactly as one whose POM a repository failed to return.

`failOnIncompleteSbom`, off by default, makes it a choice. With it on, the build stops and names every component it
could not account for, whether the dependency did not resolve or its metadata could not be read. For a fleet sharing a
remote build cache that also contains the spread: a task that fails writes nothing, so no other machine is served a
document built from a bad answer.

`allowedIncompleteComponents` keeps that switch usable in a build that depends on an artifact-only or Ivy module by
design, as full-match regular expressions over `group:name:version`. It suppresses the failure alone, so an expected
gap still appears in the document and in the log.

### Related

- #92 reports the same symptom, empty licenses for components whose license lives in a parent POM. The property would
  let that report be answered from the document instead of guessed at.
- #830 asks a neighboring question about task inputs and cache correctness.

### Contribution

Donnerbart:
Fixes #921.

Reproducer: https://github.com/Donnerbart/cyclonedx-metadata-reproducer


docs/adr/0009-report-what-the-sbom-could-not-describe.md
```diff
@@ -0,0 +1,84 @@
+---
+status: accepted
+---
+
+# Report what the SBOM could not describe
+
+A component whose POM cannot be obtained, read, or completed from its parent produced exactly the same SBOM entry as one
+that resolves and declares nothing: no licenses, no publisher, no error. `MavenProjectLookup` returned `null` from four
+separate paths, three of them without a log line, and the consumer of the document had no way to tell **Unresolved
+Metadata** from an unlicensed component. Such a component now carries a `cdx:gradle:metadata:unresolved` property naming
```
skhokhlov: This property should rather use `cdx:maven` prefix because it represents maven metadata. Also, it makes sense to register the new property [in the taxonomy](https://github.com/CycloneDX/cyclonedx-property-taxonomy/blob/main/cdx/maven.md) once it's released.

skhokhlov (changes requested):
I agree that the metadata visibility problem is worth fixing. Consumers should be able to distinguish unavailable metadata from a component that declares none. Preserving readable metadata when effective-model construction fails also makes sense.

I’d like the following changes before merging:

1. Keep this PR focused on metadata reporting. Please defer `failOnIncompleteSbom` and `allowedIncompleteComponents` to a separate proposal. They introduce a broader completeness policy that needs discussion of dependency resolution, artifact failures, metadata failures, and exceptions. Agreement with the problem described in #921 does not imply acceptance of that proposed API.
3. Avoid inferring which model input failed. The `okparentbadimport` fixture has a healthy parent and a broken imported BOM, but currently reports `parent-unparseable`. Use `model-incomplete` when effective-model construction fails without an established cause. We should not publish a more specific diagnosis than the evidence supports, especially since these values become part of the SBOM output contract.
4. Use the `cdx:maven` namespace and register the property before release. This represents Maven metadata, regardless of which build tool reports it. Please agree the property name and value meanings through the [property taxonomy](https://github.com/CycloneDX/cyclonedx-property-taxonomy/blob/main/cdx/maven.md); the [cdx rules](https://github.com/CycloneDX/cyclonedx-property-taxonomy/blob/main/cdx.md) prohibit unofficial names under that namespace. 


Keep the missing-POM behavior simple: an attempted lookup that finds no usable POM should report pom-unresolved, including intentional artifact-only or Ivy dependencies. This indicates unavailable metadata, not a defective component, and should not fail the build. No repository-specific exceptions are needed.

Please update the tests, README, glossary, and proposed ADR to match this narrower scope.

Donnerbart:
Thanks for the first review, all three are addressed.

Registration is filed as CycloneDX/cyclonedx-property-taxonomy#197, with
CycloneDX/cyclonedx-property-taxonomy#196 for the row in `cdx/maven.md`.

